PR #3, opened 2024-02-15
Stimulus optimization

### LINKED ISSUE #4: Implement MEI generation

thomasZen:
Adds code to generate MEIs. The script generated an MEI for each cell present in the model. 
Fixes https://github.com/fededagos/open-retina/issues/4.


openretina/hoefling_2022_models.py
```diff
@@ -15,154 +15,6 @@
 from .misc import set_seed
 
 
-# Batch adapation model
-def SFB3d_core_SxF3d_readout(
```
thomasZen, line 19: This whole block moved further down. I restructured it a bit to get rid of the local class.

PR #23, opened 2024-04-11
Added code to deal with new export strategy

fededagos:
This includes:
- More .h5 handling utilities
- Upsampling functions, as the new export does not contain the data upsampled to the movie frames already
- Checks as to whether the newly exported stimuli match the old ones.


openretina/neuron_data_io.py
```diff
@@ -346,3 +355,100 @@ def map_to_range(self, max, val):
...
+        upsampled_triggertimes = np.concatenate(upsampled_triggertimes)
+    else:
+        raise NotImplementedError(f"Stimulus ID {stim_id} not implemented")
+
+    # upsampled_responses = []
+    # for n, trace in enumerate(traces):
+    #     response = interp1d(tracestimes[n].flatten(), trace.flatten(), kind="linear")(upsampled_triggertimes)
+    #     upsampled_responses.append(response)
+
+    # upsampled_responses = np.stack(upsampled_responses, axis=0)
```
thomasZen: Do we still need this? If yes maybe add a comment above for what.

openretina/neuron_data_io.py
```diff
@@ -346,3 +355,100 @@ def map_to_range(self, max, val):
...
+    Args:
+        data_dict (dict): A dictionary containing the data.
+        response_type (str, optional): The type of response. Defaults to "natural".
+
+    Returns:
+        dict: The updated data dictionary with final responses.
+    Raises:
+        NotImplementedError: If the conversion is not yet implemented for the given response type.
+    """
+    stim_id = 5 if response_type == "natural" else None
```
thomasZen, line 472: When we have more stimuli we could could define a dictionary: `stim_name_to_id = {"natural": 5, ...}` and then use this as the "ground truth".

thomasZen (approved):
Cool, the docstrings are also nicely written :)

openretina/neuron_data_io.py
fededagos: You are right, I have removed it as it is not necessary anymore

openretina/neuron_data_io.py
fededagos: Fully agree! Let's add that when we have implemented all stimulus types.

PR #25, opened 2024-04-15
Incorporating chirp and moving bar stimuli into model training

fededagos:
Work in progress. So far it is not working as the length of the exported chirp and moving bar stimuli do not match my strategy for upsampling the neural responses.


openretina/neuron_data_io.py
```diff
@@ -430,16 +455,32 @@ def upsample_traces(
...
+        # Moving bar
+        #! TODO: assuming each bar has only one trigger at its start. Needs to be verified
+        bar_durations = round(np.mean(np.diff(triggertimes)))
         upsampled_triggertimes = [
-            np.linspace(t, t + ifi, up_factor, endpoint=False) for t in triggertimes
+            np.linspace(bar_t, bar_t + bar_durations, bar_durations * target_fr)
+            for bar_t in triggertimes
         ]
         upsampled_triggertimes = np.concatenate(upsampled_triggertimes)
+
```
fededagos, line 483: Here is some code I wrote to deal with the upsampling of chirp and mbs. As I wrote in the comments, I am not sure about my assumptions on the triggertimes, but the rest of the logic seems fine to me

fededagos:
I wrote [this jupyter notebook](https://github.com/fededagos/open-retina/blob/feature/training-extra-stimuli/notebooks/hoefling_2022_stimuli_viz.ipynb) to quickly visualize the stimuli (it can be deleted after we are done with this). While working in it, I noticed 2 things:

1. The `load_moving_bar_stack` function in `stimuli.py` seems to chunk up the moving bar mistakenly (it breaks in between single bar motions).
2. It is less straightforward than I assumed to align the responses to the stimuli, and for this I need help.

fededagos:
As a consequence of me not knowing how to align things properly, the function to upsample the chirp and MB responses is only tentative. My assumptions about the structure of the triggertimes need to be double checked (see comments in the code).

https://github.com/fededagos/open-retina/blob/e1786e915ac6f7e5a61e732ff2cdc137fa2c1125/openretina/neuron_data_io.py#L462C1-L482C72

openretina/hoefling_2022_data_io.py
```diff
@@ -273,3 +246,135 @@ def natmov_dataloaders_v2(
             )
 
     return dataloaders
+
+
+def chirp_dataloaders(
```
fededagos: Once the other problems are fixed, the chirp and MB dataloaders should (hopefully) work. I imagined using them by essentially merging them with the natural movies training dataloaders before passing them over to the training function. 

However, some code inside the training function (or the model) needs to change: as of now, I called the keys of the chirp and MB dataloaders with their session name followed by "_chirp" or "_mb". As it stands, the model would treat these as separate sessions. Some control flow must be implemented to ignore these suffixes.

fededagos:
Almost everything is fixed now. Again, refer to [this notebook](https://github.com/fededagos/open-retina/blob/feature/training-extra-stimuli/notebooks/hoefling_2022_stimuli_viz.ipynb). 
The only thing that needs fixing is removing one frame for each moving bar, as it is the only thing that does not match up despite all my digging. 
This is related to the fact that in the moving bars in QDSPY, each trigger is registered with a duration of 3 frames at 60 fps.
For the chirp this was negligible because there it was 2 triggers during the whole presentation, each of them presented for 1 frame at 60 fps. 
I am not sure if we can just hack away with this or need to re-export the moving bar movies.

openretina/dataloaders.py
```diff
@@ -164,3 +176,26 @@ def filter_nan_collate(batch):
     """
     batch = list(filter(lambda x: not np.isnan(x[1]).any(), batch))
     return default_collate(batch)
+
+
+def filter_different_size(batch):
+    """
+    Filters out batches that do not have the same shape as most of the other batches.
```
thomasZen, line 183: Does this happen often?
Alternatively we can use padding to make them the same size.

openretina/dataloaders.py
```diff
@@ -164,3 +176,26 @@ def filter_nan_collate(batch):
...
+    shapes = [element[1].shape for element in batch]
+
+    # Find the most common shape in the batch
+    most_common_shape = max(set(shapes), key=shapes.count)
+
+    # Filter out elements that do not have the most common shape
+    filtered_batch = [element for element in batch if element[1].shape == most_common_shape]
+
+    # If the filtered batch is empty, return None
+    if not filtered_batch:
```
thomasZen:
```suggestion
    if len(filtered_batch) == 0:
```
I prefer above as `if not filter_batch` can also trigger for None and False (and not sure how it behaves for np.array)

openretina/hoefling_2022_data_io.py
```diff
@@ -273,3 +224,132 @@ def natmov_dataloaders_v2(
...
+    dataloaders = {"train": {}}
+
+    chirp_triggers = next(iter(neuron_data_dictionary.values()))["chirp_trigger_times"][0]
+    # 2 triggers per chirp presentation
+    num_chirps = len(chirp_triggers) // 2
+
+    # Get it into chan, time, height, width
+    chirp_stimulus = torch.tensor(load_chirp(), dtype=torch.float32).permute(3, 0, 1, 2)
+
+    chirp_stimulus = chirp_stimulus.repeat(1, num_chirps, 1, 1)
```
thomasZen, line 253: The load chirp function contains an argument trigger_times. Did you try that? If so were the stimuli shapes off, too?

openretina/hoefling_2022_data_io.py
```diff
@@ -273,3 +224,132 @@ def natmov_dataloaders_v2(
...
+
+    dataloaders = {"train": {}}
+
+    mb_triggers = next(iter(neuron_data_dictionary.values()))["mb_trigger_times"][0]
+    num_repeats = len(mb_triggers) // 8
+
+    # Get it into chan, time, height, width
+    mb_stimulus = torch.tensor(load_moving_bar(), dtype=torch.float32).permute(3, 0, 1, 2)
+
+    mb_stimulus = mb_stimulus.repeat(1, num_repeats, 1, 1)
```
thomasZen, line 317: It should theoretically work to pass the trigger_times directly to load_moving_bar:
https://github.com/fededagos/open-retina/blob/50dba923e439504733833661b3e77983504825f4/openretina/stimuli.py#L93-L95

openretina/misc.py
```diff
@@ -44,6 +44,10 @@ def set_seed(seed=None, seed_torch=True):
     return seed
 
 
+class MaxLinesExceeded(Exception):
```
thomasZen:
```suggestion
class MaxLinesExceededException(Exception):
```
This makes it a bit more clear when dealing with the exception later

openretina/stimuli.py
```diff
@@ -30,6 +31,7 @@ def discretize_triggers(trigger_times: np.array) -> list[int]:
 
 
 def align_stimulus_to_trigger(stimulus: np.array, start_trigger_times: list[int]) -> np.array:
+    #! TODO: needs fixing.
```
thomasZen, line 34: What was wrong here?

openretina/stimuli.py
```diff
@@ -106,25 +108,23 @@ def load_moving_bar(
...
     moving_bar = load_moving_bar(normalize)
 
-    assert moving_bar.shape[0] % number_of_moving_bars == 0, \
-        "Moving bar timesteps are not divisible by number_of_moving_bars, something went wrong"
-    time_steps_per_bar = moving_bar.shape[0] // number_of_moving_bars
-    moving_bar_array = [moving_bar[i*time_steps_per_bar:(i+1)*time_steps_per_bar] for i in range(number_of_moving_bars)]
-    moving_bar_stack = np.stack(moving_bar_array)
+    # Remove the frames that are not part of the moving bar, i.e. black frames at the beginning and end
+    # See: https://github.com/eulerlab/QDSpy/blob/master/Stimuli/RGC_MovingBar_2.py
+    moving_bar_content = moving_bar[90:-30, ...]
```
thomasZen: Good catch

openretina/dataloaders.py
fededagos: No it does not, I put it there for something else that I was working on regarding the "cell-type-wise" training. But I agree it is an hack that should not be used if not as an exception

openretina/hoefling_2022_data_io.py
fededagos: I tried and it was lengthening the movie,

openretina/stimuli.py
fededagos: It was lengthening the movie and not matching the stim anymore, but I did not try too hard to fix it

PR #27, opened 2024-04-25
Fix nans in responses

thomasZen:
The session `session_5_ventral1_20200226_chirp` was weird. It probably had trigger times that lead to indexing in the are of Nans: 
Every session has 64 nans for the `chirp_spikes` field, 32 at the beginning and 32 at the end.

As a temporary(?) solution I ignored that particular session for the chirp.

I also added more features to `scripts/hoefling_2022_train.py`:
- You can select the datasets you want to train on
- You can specify the device type (cpu or cuda)

### DIFF
--- a/openretina/dataloaders.py
+++ b/openretina/dataloaders.py
@@ -1,5 +1,4 @@
 from collections import namedtuple
-from copy import deepcopy
 from typing import Callable, Dict, List, Optional, Tuple, TypedDict, Union
 
 import numpy as np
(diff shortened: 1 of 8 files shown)


openretina/dataloaders.py
```diff
@@ -82,7 +82,7 @@ def __len__(self):
 
 def get_movie_dataloader(
     movies: Union[np.ndarray, torch.Tensor, Dict[int, np.ndarray]],
-    responses: Float[np.ndarray, "n_neurons n_frames"],  # noqa
+    responses: Float[np.ndarray, "n_frames n_neurons"],  # noqa
```
thomasZen, line 85: I think the type annotation here was incorrect, for the chirp dataloader the shape was `n_frames n_neurons`

PR #32, opened 2024-06-06
Type checks using mypy and fix spurious type annotations, use ruff as linter and fix codestyle errors

thomasZen:
After installing the repository (do that to get the updated dependencies), run the type checks as follows:
```
./scripts/tests/codestyle.sh
```
These type checks are also automatically run now against each PR and the main branch, see the "checks" at the bottom of this PR. Codestyle.sh is using ruff.

I opted for running them against python3.10, as currently this is the lowest python version we support (we use `|` in the [code](https://github.com/open-retina/open-retina/blob/fef94cbfbed588e4189b882ec2e2c356c648788d/openretina/neuron_data_io.py#L193), if we want to support 3.9 we would have to get rid of that, but I'm in favor of only supporting >=3.10 if nobody of us uses 3.9 right now).

I added a few types, fixed a few type errors and used `type: ignore` for the remaining ones. I also fixed the codestyle of the codebase manually. In the future we can also add a linter.

Other changes:
* Small changes in `scripts/test_os_ds.py` (I can delete them again if they seem off)
* Deleted `openretina/builder.py` as its function were not used and it was a duplicate of `openretina/hoefling_2024/nnfabrik_model_loading.py`
* Fixed bug in early_stopper when maximize_objective was True (see [comment](https://github.com/open-retina/open-retina/pull/32/files#r1631284431))


openretina/early_stopping.py
```diff
@@ -133,7 +134,7 @@ def finalize(model, best_state_dict):
             if scheduler is not None:
                 scheduler.step(current_objective)
 
-            if current_objective * maximize < best_objective * maximize - tolerance:
+            if current_objective * maximize < maximize * (best_objective - tolerance):
```
thomasZen: For the maximize case this was a bug as even if the objective stayed the same, it was judged as a better one, e.g.:
```
>>> maximize_objective = True
>>> tolerance = 0.0000000000001
>>> current_objective = best_objective = 1.0
>>> 
>>> maximize = -1 if maximize_objective else 1
>>> current_objective * maximize
-1.0
>>> best_objective * maximize - tolerance
-1.0000000000001
>>> 
>>> current_objective * maximize < best_objective * maximize - tolerance
False
```

thomasZen: ^ Nevermind, I think this line was incorrect, but it is correct. I reverted my change.

scripts/tests/ruff.toml
fededagos, line 1: Good defaults!

fededagos (approved):
Everything looks clean and sensible. We can add more elaborate settings later, as you said, once we get used to having actions in the workflow.

PR #34, opened 2024-06-27
Functionality for Hoefling 2024 publication

thomasZen:
Changes to make optimization and model loading work as in the paper: [A chromatic feature detector in the retina signals visual context changes](https://www.biorxiv.org/content/10.1101/2022.11.30.518492.abstract).

## Changes:
- Early Stopper classes for stimulus optimization
- ResponseReducer classes to reduce a response trace to a single score
- Bugfix: use 0.0 instead of torch.zeros([1]) to avoid device discrepancies
- Added unit test for EarlyStopper and setup for running unit tests automatically using pytest

### DIFF
--- a/openretina/hoefling_2024/nnfabrik_model_loading.py
+++ b/openretina/hoefling_2024/nnfabrik_model_loading.py
@@ -274,8 +274,8 @@ def load_ensemble_retina_model_from_directory(directory_path: str, device: str =
     # Just put inside wrapper for ensemble
     ensemble_model = EnsembleModel(*model_list)
     # Center readouts
-    model_transform = Center(target_mean=[0.0, 0.0])
-    model_transform(ensemble_model)
+    # model_transform = Center(target_mean=[0.0, 0.0])
+    # model_transform(ensemble_model)
     ensemble_model.to(device)
     ensemble_model.eval()
 
--- a/openretina/hoefling_2024/nnfabrik_model_loading.py
+++ b/openretina/hoefling_2024/nnfabrik_model_loading.py
@@ -97,9 +97,7 @@ def get_model(
     )
 
     if state_dict is not None:
-        ignore_missing = model_config.get("transfer", False)
-        if not strict:
-            ignore_missing = True
+        ignore_missing = model_config.get("transfer", False) or not strict
         load_state_dict(
             net,
             state_dict,
@@ -266,11 +264,14 @@ def load_ensemble_retina_model_from_directory(
         data_info_list.append(data_info)
 
         model_fn = config["model_fn"]
-        [repo, _, _, model_type] = model_fn.split('.')
+        repo, _, _, model_type = model_fn.split('.')
         if repo == "nnfabrik_euler":  # convert model_fn from nnfabrik to openretina
             #ToDo check robustness across model types
-            [_, _, _, model_type] = model_fn.split('.')
             model_fn = '.'.join(['openretina', 'hoefling_2024', 'models', model_type])
+        elif repo == "openretina":
+            pass  # nothing to change
+        else:
+            raise ValueError(f"Unsupported repository {repo} for loading a model. Please implement this manually.")
         model_config = config["model_config"]
         model = get_model(model_fn, model_config, seed=seed, data_info=data_info, state_dict=state_dict, strict=False)
         model_list.append(model)
--- a/openretina/hoefling_2024/nnfabrik_model_loading.py
+++ b/openretina/hoefling_2024/nnfabrik_model_loading.py
@@ -13,6 +13,8 @@ import torch
 import torch.nn as nn
 import yaml
 
+import openretina.hoefling_2024.models
+from openretina.utils.misc import SafeLoaderWithTuple, tuple_constructor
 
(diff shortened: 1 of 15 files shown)


openretina/hoefling_2024/nnfabrik_model_loading.py
```diff
@@ -95,13 +97,16 @@ def get_model(
     )
 
     if state_dict is not None:
+        ignore_missing = model_config.get("transfer", False)
+        if not strict:
+            ignore_missing = True
```
thomasZen: Could write that shorter
```suggestion
        ignore_missing = model_config.get("transfer", False) or not strict
```

openretina/hoefling_2024/nnfabrik_model_loading.py
```diff
@@ -250,22 +260,26 @@ def load_ensemble_retina_model_from_directory(directory_path: str, device: str =
...
         with open(model_config_path, "r") as f:
-            config = yaml.safe_load(f)
-
+            config = yaml.load(f, SafeLoaderWithTuple)
         with open(data_info_path, "rb") as fb:
             data_info = pickle.load(fb)
         data_info_list.append(data_info)
 
         model_fn = config["model_fn"]
+        [repo, _, _, model_type] = model_fn.split('.')
```
thomasZen: parenthesis are not required.
```suggestion
        repo, _, _, model_type = model_fn.split('.')
```

openretina/hoefling_2024/nnfabrik_model_loading.py
```diff
@@ -250,22 +260,26 @@ def load_ensemble_retina_model_from_directory(directory_path: str, device: str =
...
+        if repo == "nnfabrik_euler":  # convert model_fn from nnfabrik to openretina
+            #ToDo check robustness across model types
```
thomasZen, line 269: Do we still have to check this?

openretina/hoefling_2024/nnfabrik_model_loading.py
```diff
@@ -250,22 +260,26 @@ def load_ensemble_retina_model_from_directory(directory_path: str, device: str =
...
+            [_, _, _, model_type] = model_fn.split('.')
```
thomasZen: Do we need this line? It seem the line just sets model_type to the same value as it was already set by the statement of line 269:
```
[repo, _, _, model_type] = model_fn.split('.')
```

thomasZen (commented):
Added a few comments.
Somebody else could ideally review the rest of the code as I implemented the other changes.

openretina/hoefling_2024/nnfabrik_model_loading.py
lhoefling: You think this will always be nnfabrik_euler?

thomasZen: > You think this will always be nnfabrik_euler?

Maybe we could do the following:
```
if repo == "nnfabrik_euler":
   ...
elif repo == "openretina":
    pass  # nothing to change
else:
    raise ValueError(f"Unsupported repository {repo} for loading a model. Please implement this manually.")
```
Then, it's clear to the user that he has to look at the code to figure out how to load his model.

PS: Somehow I only saw your comment via mail and not in Github, that's why I pasted the comment from the mail.

openretina/optimization/regularizer.py
```diff
@@ -1,67 +1,55 @@
...
-    "x_max_green": 6.269,
-    "x_min_uv": -0.913,
-    "x_max_uv": 6.269,
-}
-"""
-
+class RangeRegularizationLoss(StimulusRegularizationLoss):
+    def __init__(
+            self,
+            min_max_values: Iterable[tuple[float, float]] = ((-0.654, 6.269), (-0.913, 6.269)),
```
fededagos: Could these defaults be added to hoefling_2024/constants maybe?

thomasZen: Good point. I added them to hoefling_2024/constants and removed the defaults here.

openretina/optimization/regularizer.py
```diff
@@ -1,67 +1,55 @@
...
-        max_penalty = torch.sum(torch.relu(stimulus_i - max_val))
-        penalty += min_penalty + max_penalty
+    def forward(self, stimulus: torch.Tensor) -> torch.Tensor:
+        """ Penalizes the stimulus if it is outside the range defined by min_max_values. """
+        loss: torch.Tensor = 0.0  # type: ignore
+        for i, (min_val, max_val) in enumerate(self._min_max_values):
+            stimulus_i = stimulus[:, i]
+            min_penalty = torch.sum(torch.relu(min_val - stimulus_i))
+            max_penalty = torch.sum(torch.relu(stimulus_i - max_val))
+            loss += min_penalty + max_penalty
```
lhoefling, line 31: This does not necessarily force the stimulus to be in range, right?

lhoefling: I see we take care of that with postprocessing, so my comment is obsolete

lhoefling (approved):

PR #35, opened 2024-06-28
Model learning readout weights from each cell metadata

fededagos:
This PR tries to add a new model class which uses a "frozen" readout, where the spatial weights come from other learned readouts on the same data (e.g. trained on the moving bar), and the feature weights are computed from a sub-network that uses each cell's metadata.


openretina/dataloaders.py
```diff
@@ -203,3 +328,20 @@ def filter_different_size(batch):
...
+
+    Args:
+        batch (list): A list of tuples representing the batch.
+
+    Returns:
+        tuple of torch.Tensor: The collated batch after filtering out empty videos.
+
+    """
+    # Filter out empty videos
+    batch = list(filter(lambda x: x[0].shape[1] > 0, batch))
```
thomasZen: I think python folks prefer list comprehensions over the "functional filter style". Just for information, if you like filter keep it.

```suggestion
    batch = [x for x in batch if x[0].shape[1] > 0]
```

openretina/dev_models.py
```diff
@@ -590,3 +724,70 @@ def __init__(self, in_shape_dict, n_neurons_dict, **kwargs):
...
+        self.outdims = outdims
+        self.init_noise = init_noise
+        c, w, h = in_shape
+
+        self.linear = torch.nn.Linear(in_features=c * w * h, out_features=outdims, bias=False)
+        if bias:
+            bias = torch.nn.Parameter(torch.Tensor(outdims))
+            self.register_parameter("bias", bias)
+        else:
+            self.register_parameter("bias", None)
```
thomasZen, line 758: I was struggling how to deal with None values. Is this the correct way (I think I read up on register_parameters, but not as much as to make a decision)

openretina/dev_models.py
```diff
@@ -590,3 +724,70 @@ def __init__(self, in_shape_dict, n_neurons_dict, **kwargs):
...
+
+        self.initialize()
+
+    @property
+    def features(self):
+        return next(iter(self.linear.parameters()))
```
thomasZen: Should this return the weight matrix of the linear layer? If so I think you should do:
```suggestion
        return self.linear.weight
```

openretina/dev_models.py
thomasZen: Also why not use just use the bias vector of torch.nn.Linear: ` self.linear = torch.nn.Linear(in_features=c * w * h, out_features=outdims, bias=bias)`?

openretina/hoefling_2024/models.py
```diff
@@ -126,11 +122,11 @@ def __init__(
         else:
             hidden_pad = [0 for _ in range(1, len(spatial_kernel_size))]
 
-        if not isinstance(hidden_channels, Iterable):
+        if not isinstance(hidden_channels, (list, tuple)):
```
thomasZen, line 171: Why are these changes necessary / What broke without them?

openretina/neuron_data_io.py
```diff
@@ -239,8 +241,10 @@ def __init__(
         self.eye = eye if eye is not None else "right"
         self.group_assignment = group_assignment
         self.key = key
-        self.roi_coords = roi_coords
         self.roi_ids = roi_ids
+        self.roi_coords = (
+            torch.tensor(self.transform_roi_mask(roi_mask), dtype=torch.float32) if roi_mask is not None else None
+        )
```
thomasZen: Would this already break Lara's code to reproduce the paper?

openretina/neuron_data_io.py
```diff
@@ -333,7 +310,49 @@ def response_dict(self):
             },
         }
 
-        return response_dict
+    def compute_validation_responses(self):
```
thomasZen:
```suggestion
    def compute_validation_responses(self) -> None:
```
This is sometimes good to add because the type checker does not check functions that have no type annotations in their signature

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+from jaxtyping import Float
+from neuralpredictors.layers.readouts import MultiReadoutBase, Readout
+from scipy.fftpack import fft
+from torch.utils.data import DataLoader
+from tqdm.auto import tqdm
+
+from .dataloaders import MovieAndMetadataDataSet, MovieSampler
+from .dev_models import (  # ! TODO: move the model to dev models once done developing.
```
thomasZen, line 17: Maybe we should make a folder called `dev_models`, the `dev_models.py` file is already quite big.

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+        arithmetic_mean = np.mean(magnitudes)
+        return geometric_mean / (arithmetic_mean + np.finfo(float).eps)
+
+    spectral_centroid_value = np.apply_along_axis(spectral_centroid, 1, chirp_trace)
+    spectral_flatness_value = np.apply_along_axis(spectral_flatness, 1, chirp_trace)
+    mean_value = np.mean(chirp_trace, axis=1)
+    std_value = np.std(chirp_trace, axis=1)
+    range_value = np.ptp(chirp_trace, axis=1)
+
+    return np.stack([spectral_centroid_value, spectral_flatness_value, mean_value, std_value, range_value])
```
thomasZen, line 213: Any insights about these features (e.g. why did you choose them)

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+
+class ReadoutWeightShifter(nn.Module):
+    def __init__(self, num_numerical_features, categorical_vocab_sizes, categorical_embedding_dims, output_dim):
+        super(ReadoutWeightShifter, self).__init__()
+
+        # Embedding layers for categorical features
+        self.embeddings = nn.ModuleList(
+            [
+                nn.Embedding(vocab_size, categorical_embedding_dim)
+                for vocab_size, categorical_embedding_dim in zip(categorical_vocab_sizes, categorical_embedding_dims)
```
thomasZen:
```suggestion
                for vocab_size, categorical_embedding_dim in zip(categorical_vocab_sizes, categorical_embedding_dims, strict=True)
```
This is good to prevent errors as it checks that both sequences have the same length and throws an error otherwise.

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+        dataset = MovieAndMetadataDataSet(movies, responses, metadata, split, chunk_size)
+        sampler = MovieSampler(
+            start_indices, split, chunk_size, movie_length=movies.shape[1], scene_length=scene_length
+        )
+
+    return DataLoader(
+        dataset,
+        sampler=sampler,
+        batch_size=batch_size,
+        drop_last=True if (split == "train" and drop_last) else False,
```
thomasZen: Maybe this was copied, but you can just write:
```suggestion
        drop_last=split == "train" and drop_last,
```

I'm also much against having something like:
```
x = True
y = True if x else False
```
in the codebase :)

openretina/cell_type_training.py
thomasZen: (I assume you use these features/barcodes as input to the model afterwards):
If this works great, but I'm skeptical if hand designing features will work without a lot of experimentation.
My thought was to define a small network (probably just a small LSTM or so) and throw in the avg trace over the chirp and/or moving bar responses.
This network then outputs the barcode features for a cell, and the network can be trained end to end.

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+            traces=traces,
+            tracestimes=tracestimes,
+            scan_sequence_idx=scan_sequence_idx,
+            random_sequences=random_sequences,
+            eye=eye,
+            group_assignment=group_assignment,
+            key=key,
+            use_base_sequence=use_base_sequence,
+            **kwargs,
+        )
```
thomasZen: Is there something missing here? Shouldn't the barcode be stored to?

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+        pdf = torch.sum(pdf**2, dim=-1) / (self.mask_var_ + 1e-8)
+        pdf = torch.exp(-0.5 * torch.clamp(pdf, max=20))
+        normalisation = torch.sum(pdf, dim=(1, 2), keepdim=True)
+        pdf = torch.nan_to_num(pdf / normalisation)
+        return pdf
+
+    def forward(self, x, features, subs_idx=None):
+        b, c, w, h = x.size()
+
+        features = features.view(-1, c, self.outdims)
```
thomasZen: I'm unsure if this view is correct (although I don't fully understand the code). The `-1` dimension is normally the batch dimension, and it might result in a weird bug to mix up features in the batch dimension.

Or can this be changed to `features = features.view(b, c, self.outdims)`?

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+        if subs_idx is not None:
+            feat = features[..., subs_idx]
+            masks = self._masks[..., subs_idx]
+
+        else:
+            feat = features
+            masks = self._masks
+
+        y = torch.einsum("ncwh,whd->ncd", x, masks)
+        y = (y * feat).sum(1)
```
thomasZen: To check my understanding: I assume this is the same code as before, just with features not a learnable parameters of this modules but calculated by the shifter network

openretina/cell_type_training.py
thomasZen: Maybe this view call should even happen in ShifterVideoEncoder, near the comment `# Treat time as an indipendent (batch) dimension for the readout`. Then it's clear that both x and features are viewed in that way.

openretina/cell_type_training.py
```diff
@@ -0,0 +1,697 @@
...
+        # Transpose after concatenating to have feature dimension as second
+        embedded_cats = torch.cat(embedded_cats, dim=1).transpose(2, 1)
+
+        # Concatenate numerical and embedded categorical features, put feature dimension as last to prepare forward
+        x = torch.cat([numerical_input, embedded_cats], dim=1).transpose(1, 2)
+
+        # Forward pass, then put features back as second dimension for batch norm
+        x = self.fc1(x).transpose(1, 2)
+        # Transpose again after batchnorm to have feature dimension as last again
+        x = F.relu(self.bn1(x)).transpose(1, 2)
```
thomasZen: Having so many transpose calls seems to be error prone. Maybe simplify or check if the dimensions are really still right (one way to check is to use random inputs for which each dimension is a different prime number (e.g. inp = torch.rand((2, 3, 5, 7)) and then checking the shape of each tensor in a debugger.)

thomasZen (commented):
I didn't find any obvious errors, but commented on some parts of the code where I think bugs might be more likely.

It also might be worth a try to just feed the chirp and/or moving bar features into the ReadoutWeightShifter instead of feeding hand designed features into it.

openretina/cell_type_training.py
fededagos: Agree, should definitely be done before this is merged (would also be useful going forward in development)

openretina/cell_type_training.py
fededagos: I chose these features as they are easily computable from the chirp and MB traces (some are directly exported from the data pipeline), and most importantly they are human interpretable. They are absolutely not a complete set, and I am totally open to discussion on whether some should be dropped or some more added. The idea would be to be able, ultimately once the model is trained and works, to specify which kind of cell you want to simulate the output of, and having the features of that shifter network human-interpretable would allow an experimenter to say what they want to simulate with a good amount of control. 
On the other hand I also agree that DL usually works better with continuous high-dim data. I would be totally up to use just the full trace to get a better idea of whether this could work or not.

openretina/cell_type_training.py
fededagos: Great suggestion

openretina/cell_type_training.py
fededagos: Will add as todo. 
I agree there could be a cleaner way to do this...

openretina/cell_type_training.py
fededagos: Will check on this. -1 is indeed batch there, and that could already be changed, but making it happen at the end of the shifter is also a good idea

openretina/cell_type_training.py
fededagos: This is indeed right. The masks are also not learned but should be imported from another network (e.g. a network trained with a factorised readout on the MB responses, or something else that allows learning the "spatial receptive fields" in the core feature space)

openretina/cell_type_training.py
fededagos: Something is indeed missing right now... I just ended up passing the barcode elsewhere, so this should either be removed or the barcode added here. Will update

openretina/dev_models.py
fededagos: I have to admit I do not know how to answer this. This initial definition is taken straight from the neuralpredictors one for the factorised readout. In this case we could indeed just use the bias from nn.Linear .

openretina/hoefling_2024/models.py
fededagos: PyLance would complain of wrong typing :)
(there was some operation done on those elements that was not compatible with any iterable but only with list and tuple, which is what is only appropriate to use there anyway)

openretina/neuron_data_io.py
fededagos: I am not sure about this but there are definitely other things in this PR that might break the reproduction code. We should probably merge this to main 1. If it works 2. after Lara is done with all reproduction code and can point to a safe branch that we will keep for that purpose.

openretina/hoefling_2024/models.py
thomasZen: Ah, makes sense.
It's not relevant for the functionality right now, but in case we want to merge this we'd have to double check though as I think currently the code wouldn't work with Iterables I think, as the next line is `hidden_channels = [hidden_channels] * (self.layers)` and:
```
[iter(range(10))] * (10)
[<range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>, <range_iterator object at 0x7ec8d83679c0>]
````
I think the code should be something like:
```
if isinstance(hidden_channels, Iterable):
    hidden_channels_list = list(hidden_channels)
else:
    hidden_channels_list = [hidden_channels] * self.layers
```
(or we just only support that the argument hidden_channels is a list)

File: openretina/neuron_data_io.py (hunk 5 above)
Comment (thomasZen) [reply openretina/neuron_data_io.py]: Yes, sounds good.

Comment (thomasZen) [reply openretina/neuron_data_io.py]: Maybe we can additionally add simplified tests for the main functionality for Lara's code to make sure they keep working.

Comment (fededagos) [reply openretina/neuron_data_io.py]: Yes I think it would be good to code up some simple tests to check for example that a model object can still be initialised with some dummy data and training can be started

**Comment by fededagos** [comment]
Training stabilised with some minor tweaks:

- Added tanh at the end of `ReadoutWeightShifter` so that it can only positively and negatively weigh the feature outputs between -1 and +1
- Double-checked feature normalisation (the option was there but not applied in the training script I was using)
- Applied gradient clipping to the trainer
- Slightly played around with the learning rate: a lr of 0.05 seems to make the model train to 0.2 test accuracy

Now further experiments are needed to see how high the accuracy can get. Ideally the final evaluation should be done on a completely unseen session, as this is something we can afford with this new model type.

File: openretina/cell_type_training.py (hunk 12 above)
Comment (fededagos) [reply openretina/cell_type_training.py]: Double checked. Was leading to a bug: not for the -1 as batch, but for the reshape instead of transpose of the channel and output dimension, which is now done before this call.

File: .github/workflows/python-tests.yml [hunk 15]
```diff
@@ -28,13 +30,19 @@ jobs:
...
         run: |
           ./scripts/tests/codestyle.sh
+
       - name: Type checks
+        if: matrix.task == 'type-checks'
         run: |
           ./scripts/tests/check-types.sh
+
       - name: Units tests
+        if: matrix.task == 'unit-tests'
```
thomasZen: These changes force the CI to always run all checks, right?
I'd prefer to directly add this change in a separate PR, as it isn't related to the other code.

fededagos: I agree, sorry for adding it here. Will remove and open a separate PR

fededagos:
In [4383992](https://github.com/open-retina/open-retina/pull/35/commits/438399270a77e3350a69576cd9fa4358e5f37fd9) I have added a notebook to play around with the model. For it to work, you have to extract and put into openretina/models the model files that you can download [here](https://files.fededagos.me/18-10-2024_models.zip), and modify the paths in the notebook accordingly.

PR #36, opened 2024-07-01
Changes to upload repository to pypi

thomasZen:
I followed this tutorial: https://packaging.python.org/en/latest/tutorials/packaging-projects/
To build and upload the package you can use the following commands (provided that you installed twine using pip and you got access to the pypi api token):
```
python3 -m build
twine upload dist/*
```

# Changes
- Updated with authors and project urls pointing to this github page
- Got rid of the neuralpredictors dependency. This is necessary because for publishing a python package on pypi you must not have dependencies that are install from github based on commit hashes. I kept the import from neuralpredictors for the dev_models and used type ignores for them.
- Include data pickle files of chirp and moving bar stimuli in package following [this setuptools guide](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files)

## Installing the open-retina package
This currently works from the test pypi server using the following command:
```
pip install --index-url https://test.pypi.org/simple/ --no-deps openretina
```
Once it's uploaded to the actual pypi server we can use `pip install openretina`.


fededagos:
Good idea. It would be a nice way to point to the specific version of the package that can reproduce the paper reliably and have it pip-installable more easily

fededagos (approved):
I have no suggestions to make at this stage: readme looks good to get started, and indeed the functions left to move from neuralpredictors are few and minor. We have more in dev_models but we can figure that out going forward.

PR #37, opened 2024-07-08
Option to use a readout rnn for each session

thomasZen:
The readout rnn is a GRU unit. To activate it set `use_readout_rnn` to True in the in the config.


.gitignore
```diff
@@ -166,5 +166,5 @@ data/
 random.ipynb
 *.jpg
 *.pdf
-figures/
-models/
+/figures
+/models
```
thomasZen: This only ignores these folder in the root directory of the github repository, and not all directories with that name.

openretina/hoefling_2024/models.py
```diff
@@ -446,8 +460,13 @@ def forward(self, x, shift=None, subs_idx=None):
             feat = self.features
             masks = self.masks
 
-        y = torch.einsum("nctwh,whd->nctd", x, masks)
-        y = (y * feat).sum(1)
+        y_einsum = torch.einsum("nctwh,whd->nctd", x, masks)
+
+        if self._readout_rnn is not None:
+            y = torch.permute(y_einsum[:, :, :, 0], (0, 2, 1))
```
fededagos: This is still working only for one neuron right? How did you run it on all? I suppose putting neuron and batch dimension together and then reshaping again after the rnn is done could work here. Also, why go fully connected in the rnn? Did you try keeping the feature re-weighting?

thomasZen:
Closing as it's outdated.

PR #38, opened 2024-07-08
Cleanup and type annotations

thomasZen:
- Remove a few unnecessary functions and arguments
- Added type annotations
- Fixed bug (see comments)

I will test if it's still running with Lara's model.

### DIFF
--- a/openretina/hoefling_2024/models.py
+++ b/openretina/hoefling_2024/models.py
@@ -1,6 +1,6 @@
 from collections import OrderedDict
 from collections.abc import Iterable
-from typing import Literal, Optional, Tuple
+from typing import Literal, Optional, Tuple, Callable
 
 import numpy as np
 import torch
@@ -31,16 +31,11 @@ class Core:
 
 
 class Core3d(Core):
-    def initialize(self, cuda=False):
+    def initialize(self):
         self.apply(self.init_conv)
-        self.put_to_cuda(cuda=cuda)
-
-    def put_to_cuda(self, cuda):
-        if cuda:
-            self = self.cuda()
 
     @staticmethod
-    def init_conv(m):
+    def init_conv(m) -> None:
         if isinstance(m, nn.Conv3d):
             nn.init.xavier_normal_(m.weight.data)
             if m.bias is not None:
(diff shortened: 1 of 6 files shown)


openretina/hoefling_2024/models.py
```diff
@@ -31,16 +31,11 @@ def __repr__(self):
...
 
 class Core3d(Core):
-    def initialize(self, cuda=False):
+    def initialize(self):
         self.apply(self.init_conv)
-        self.put_to_cuda(cuda=cuda)
-
-    def put_to_cuda(self, cuda):
-        if cuda:
-            self = self.cuda()
```
thomasZen, line 33: This happens automatically if we call model.to(device)

openretina/optimization/objective.py
```diff
@@ -4,7 +4,7 @@
 
 class ResponseReducer(ABC):
     def __init__(self, axis: int = 0):
-        self._axis = 0
+        self._axis = axis
```
thomasZen, line 7: This was a bug

openretina/training.py
```diff
@@ -149,8 +149,8 @@ def full_objective(model, data_key, inputs, targets, detach_core):
             if (batch_no + 1) % optim_step_count == 0:
                 optimizer.step()
                 optimizer.zero_grad()
-        if np.isnan(loss.item()):
-            raise ValueError(f"Loss is NaN on batch {batch_no} from {data_key}, stopping training.")
+            if np.isnan(loss.item()):
+                raise ValueError(f"Loss is NaN on batch {batch_no} from {data_key}, stopping training.")
```
thomasZen, line 153: This was also a bug, as we only raised the ValueError if the last loss was a nan, but we want to raise it if any loss was nan.

openretina/hoefling_2024/models.py
fededagos: Did you test this? I agree that having a put_to_cuda method does not make sense, but it seems that Core is not inheriting from nn.Module right now so .to(device) might not work

openretina/training.py
fededagos: Good catch!

openretina/hoefling_2024/models.py
fededagos: But given that I have never used that method myself and only .to(device) I think it should work already

thomasZen: You're right that model.to(device) doesn't work as the class doesn't inherit from nn.Module.
For the same reason .to(device) doesn't work:
```
>>> from openretina.hoefling_2024.models import Core3d
>>> c3d = Core3d()
>>> c3d.cuda()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Core3d' object has no attribute 'cuda'
```
I think actually the put_to_cuda function was never actively used.

I will go ahead and let Core inherit from nn.Module, but still get rid of the push_to_cuda method.

fededagos (approved):
Looks all good to me!

PR #40, opened 2024-07-19
Functionalities for Model Visualization

thomasZen:
## New Functionality
- Create MEIs for internal model neurons
- Sparse Autoencoder
- Visualize Weights of `SpatialXFeature3d` and `STSeparableBatchConv3d` layer
- Add ContrastiveNeuronObjective of https://github.com/ecker-lab/most-discriminative-stimuli

### SpatialXFeature3d Readout Layer Weights Visualization
neuron_2


scripts/visualize_model_neurons.py
```diff
@@ -0,0 +1,184 @@
...
+    ChangeNormJointlyClipRangeSeparately,
+    RangeRegularizationLoss,
+)
+from openretina.plotting import plot_stimulus_composition
+from openretina.hoefling_2024.nnfabrik_model_loading import load_ensemble_retina_model_from_directory, Center
+
+
+BASE_PATH = "/gpfs01/euler/data/SharedFiles/projects/Hoefling2024/"
+BASE_PATH = "/home/tzenkel/GitRepos/rgc-natstim-model/data/"
+ENSEMBLE_MODEL_PATH = BASE_PATH + "models/nonlinear/9d574ab9fcb85e8251639080c8d402b7"
```
fededagos:
```suggestion
ENSEMBLE_MODEL_PATH = os.path.join(BASE_PATH, "models/nonlinear/9d574ab9fcb85e8251639080c8d402b7")
```

scripts/visualize_model_neurons.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="")
+
+    parser.add_argument("--save_folder", type=str, help="Path were to save outputs", default=".")
+    parser.add_argument("--device", type=str, choices=["cuda", "cpu"], default="cuda")
+    parser.add_argument("--model_id", type=int, default=1)
```
fededagos: Could add the ensemble path as an argument, and leave the constants defined above as a default for faster experimentation

scripts/visualize_model_neurons.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def load_model(
+        path: str = ENSEMBLE_MODEL_PATH,
+        device: str = "cuda",
+        model_id: int = 0,
+        do_center_readout: bool = False
+):
+    center_readout = Center(target_mean=(0.0, 0.0)) if do_center_readout else None
+    data_info, ensemble_model = load_ensemble_retina_model_from_directory(
```
fededagos, line 46: Right now this will only work on models saved with Lara's procedure. This made me realise that our saving functions are incomplete, as they only save the model weights and not 1. The data info and 2. the model config.
We should change our saving function and then take the `load_ensemble_retina_model_from_directory` to a potential `model_io.py` file where we also have something to load single models. What do you think?

scripts/visualize_model_neurons.py
```diff
@@ -0,0 +1,184 @@
...
+    )
+    optimizer_init_fn = partial(torch.optim.SGD, lr=100.0)
+
+    data_key = model.readout_keys()[0]
+    inner_neuron_objective = InnerNeuronVisualizationObjective(model, data_key, response_reducer)
+    layer_names_array = [x for x in inner_neuron_objective.features_dict.keys()
+                         if "readout" not in x and "regularizer" not in x and x != "core_features"]
+    print(layer_names_array)
+    for layer_name in layer_names_array:
+        for channel_id in range(16):
```
fededagos: Could simplify the nested loop using itertools.product. Also, why the hardcoded 16?

scripts/visualize_model_neurons.py
```diff
@@ -0,0 +1,184 @@
...
+            print(f"Optimizing {layer_name=} {channel_id=}")
+            stimulus = torch.randn(stimulus_shape, requires_grad=True, device=device)
+            stimulus.data = stimulus_postprocessor.process(stimulus.data)
+            inner_neuron_objective.set_layer_channel(layer_name, channel_id)
+            if "layer0" in layer_name:
+                response_reducer._start = 20
```
fededagos: Could you add a comment to explain this assignment?

scripts/visualize_model_neurons.py
thomasZen: Done.

scripts/visualize_model_neurons.py
thomasZen: I think for all new models we should rely on pytorch lightning.
E.g. they have a function called [save_hyperparameters](https://lightning.ai/docs/pytorch/stable/common/lightning_module.html#id2) which should cover 2. And for future models they should not rely on the data dict directly, but maybe just on a list of data_keys?

For the current models I agree. How about implementing a static/class method for the Ensemble model called [load_from_checkpoint](https://lightning.ai/docs/pytorch/stable/common/lightning_module.html#id3) that has the same interface as the pytorch lightning function?

I would be in favor of doing that in a separate PR, though.

scripts/visualize_model_neurons.py
thomasZen: Good point, I changed the code to figure out the number of channels dynamically. Unfortunately, it got more challenging to switch to itertools.product after this change, though I kept the nested loop.

scripts/visualize_model_neurons.py
thomasZen: I moved and simlified this code, and added a comment.

fededagos (approved):
All looking good.

PR #41, opened 2024-07-26
Polarplot for direction selectivity

thomasZen:


openretina/metrics.py
```diff
@@ -38,11 +38,11 @@ def model_predictions(loader, model: torch.nn.Module, data_key, device):
         #             output = torch.cat((output, curr_output.detach().cpu()), dim=0)
         output = torch.cat((output, (model(images.to(device), data_key=data_key).detach().cpu())), dim=0)
         target = torch.cat((target, responses.detach().cpu()), dim=0)
-    output = output.numpy()
-    target = target.numpy()
-    lag = target.shape[1] - output.shape[1]
+    output_np = output.numpy()
+    target_np = target.numpy()
```
thomasZen: Type checks now complain about when output is first a `torch.Tensor` and then gets assigned to a `numpy.ndarray`. Making the names a bit more explicit fixed that.

fededagos (approved):
Just realised we forgot about this pr. All looks good to me, you can go on and merge!

openretina/metrics.py
thomasZen: For context, I think this only happens for numpy 2.0

PR #44, opened 2024-08-02
Dev: Run model on different rgc subsets

thomasZen:
Only for sharing code, not for merging.


thomasZen:
Changes today:
- Fixed off by one error when generating mask of relevant_cells (ids start at 1 and not at 0)
- Also plot chirp responses

thomasZen:
Closing as it's outdated and currently not needed.

PR #46, opened 2024-08-15
Quality of life changes from feature/cell-type-training

fededagos:
- Added einops to requirements
- Modified .gitignore for flexibility
- Added more schedulers and optimizers options to trainer
- Option to filter data by quality criteria in IO functions
- Miscellaneous helper functions
- Ignoring Ruff errors related to jaxtyping

### DIFF
--- a/openretina/metrics.py
+++ b/openretina/metrics.py
@@ -15,7 +15,8 @@ def correlation_numpy(
     """Compute the correlation between two NumPy arrays along the specified dimension(s)."""
     y1 = (y1 - y1.mean(axis=axis, keepdims=True)) / (y1.std(axis=axis, keepdims=True, ddof=0) + eps)
     y2 = (y2 - y2.mean(axis=axis, keepdims=True)) / (y2.std(axis=axis, keepdims=True, ddof=0) + eps)
-    return (y1 * y2).mean(axis=axis, **kwargs)
+    corr = (y1 * y2).mean(axis=axis, **kwargs)
+    return corr
 
 
 def MSE_numpy(y1: np.ndarray, y2: np.ndarray, axis: None | int | tuple[int] = -1, **kwargs) -> np.ndarray:
@@ -72,7 +73,7 @@ def corr_stop(model: torch.nn.Module, loader, avg: bool = True, device: str = "c
         ret = correlation_numpy(target, output, axis=0)
 
         if np.any(np.isnan(ret)):
-            warnings.warn("{}% NaNs ".format(np.isnan(ret).mean() * 100))
+            warnings.warn(f"{np.isnan(ret).mean() * 100}% NaNs ")
         ret[np.isnan(ret)] = 0
 
         if not avg:
@@ -106,7 +107,7 @@ def corr_stop3d(model: torch.nn.Module, loader, avg: bool = True, device: str = 
         ret = ret.mean(axis=0)
 
         if np.any(np.isnan(ret)):
-            warnings.warn("{}% NaNs in corr_stop3d".format(np.isnan(ret).mean() * 100))
+            warnings.warn(f"{np.isnan(ret).mean() * 100}% NaNs in corr_stop3d")
         ret[np.isnan(ret)] = 0
 
         if not avg:
@@ -128,7 +129,7 @@ def poisson_stop(model: torch.nn.Module, loader, avg: bool = False, device: str 
 
         ret = output - target * np.log(output + EPSILON)
         if np.any(np.isnan(ret)):
-            warnings.warn(" {}% NaNs ".format(np.isnan(ret).mean() * 100))
+            warnings.warn(f" {np.isnan(ret).mean() * 100}% NaNs ")
 
         poisson_losses = np.append(poisson_losses, np.nansum(ret, 0))
         n_neurons += output.shape[1]
@@ -223,8 +224,8 @@ def compute_oracle(responses, predictions=None) -> tuple[np.ndarray, np.ndarray]
     :param predictions: array of shape time x #cells
     :return: array of shape #cells containing oracle scores
     """
+    n_cells, _, n_reps = responses.shape
     if predictions is None:
-        n_cells, _, n_reps = responses.shape
         oracle = np.zeros_like(responses)
         oracle_score = np.zeros(n_cells)
         for cell in range(n_cells):
@@ -241,7 +242,6 @@ def compute_oracle(responses, predictions=None) -> tuple[np.ndarray, np.ndarray]
             oracle_score[cell] = correlation_numpy(x, y)
         return oracle, oracle_score
     else:
-        n_cells, _, n_reps = responses.shape
         oracle_score = np.zeros(n_cells)
         for cell in range(n_cells):
             x = np.tile(predictions[:, cell], n_reps)
--- a/openretina/metrics.py
+++ b/openretina/metrics.py
@@ -6,20 +6,28 @@ import torch
 from openretina.models.model_utils import eval_state
 from openretina.utils.constants import EPSILON
 
+from .utils.misc import tensors_to_device
+
 
 def correlation_numpy(
-        y1: np.ndarray,
-        y2: np.ndarray,
-        axis: None | int | tuple[int] = -1,
-        eps: float = 1e-8,
-        **kwargs
+    y1: np.ndarray, y2: np.ndarray, axis: None | int | tuple[int] = -1, eps: float = 1e-8, **kwargs
 ) -> np.ndarray:
-    """ Compute the correlation between two NumPy arrays along the specified dimension(s). """
+    """Compute the correlation between two NumPy arrays along the specified dimension(s)."""
     y1 = (y1 - y1.mean(axis=axis, keepdims=True)) / (y1.std(axis=axis, keepdims=True, ddof=0) + eps)
     y2 = (y2 - y2.mean(axis=axis, keepdims=True)) / (y2.std(axis=axis, keepdims=True, ddof=0) + eps)
-    corr = (y1 * y2).mean(axis=axis, **kwargs)
+    return (y1 * y2).mean(axis=axis, **kwargs)
 
-    return corr
+
+def MSE_numpy(y1: np.ndarray, y2: np.ndarray, axis: None | int | tuple[int] = -1, **kwargs) -> np.ndarray:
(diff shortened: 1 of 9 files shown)


openretina/metrics.py
```diff
@@ -6,20 +6,28 @@
...
 ) -> np.ndarray:
-    """ Compute the correlation between two NumPy arrays along the specified dimension(s). """
+    """Compute the correlation between two NumPy arrays along the specified dimension(s)."""
     y1 = (y1 - y1.mean(axis=axis, keepdims=True)) / (y1.std(axis=axis, keepdims=True, ddof=0) + eps)
     y2 = (y2 - y2.mean(axis=axis, keepdims=True)) / (y2.std(axis=axis, keepdims=True, ddof=0) + eps)
-    corr = (y1 * y2).mean(axis=axis, **kwargs)
+    return (y1 * y2).mean(axis=axis, **kwargs)
+
 
-    return corr
```
thomasZen: I like the a bit more verbose style more as it's easier for debugging.
e.g. you can set a breakpoint right before the return statement and can check the correlation.

openretina/utils/misc.py
```diff
@@ -124,9 +144,69 @@ def redirect_stdout(file=open(os.devnull, "w")):
...
+
 def tuple_constructor(loader, node):
     return tuple(loader.construct_sequence(node))
 
 
 class SafeLoaderWithTuple(yaml.SafeLoader):
     pass
+
+
+class CaptureOutputAndWarnings:
```
thomasZen, line 171: Why is this needed?

openretina/training.py
```diff
@@ -195,15 +238,15 @@ def save_checkpoint(model, optimizer, epoch, loss, save_folder: str, model_name:
 
 
 def save_model(model: torch.nn.Module, save_folder: str, model_name: str) -> None:
-    os.makedirs(save_folder, exist_ok=True)
+    if not os.path.exists(save_folder):
+        # only create the lower level directory if it does not exist
```
thomasZen, line 242: Sorry for changing this previously, I didn't know that was the intention.

openretina/training.py
```diff
@@ -143,26 +162,50 @@ def full_objective(model, data_key, inputs, targets, detach_core):
             leave=True,
             disable=not verbose,
         ):
-            clean_data_key = clean_session_key(data_key)
-            loss = full_objective(model, clean_data_key, *data, detach_core)  # type: ignore
+            # clean the data key to use the same readout if we are training on multiple stimuli
+            clean_data_key = clean_session_key(data_key) if multiple_stimuli else data_key
```
thomasZen, line 166: good catch

thomasZen (commented):
Added a few comments, will read the rest tomorrow.

openretina/metrics.py
```diff
@@ -6,20 +6,28 @@
...
 def correlation_numpy(
-        y1: np.ndarray,
-        y2: np.ndarray,
-        axis: None | int | tuple[int] = -1,
-        eps: float = 1e-8,
-        **kwargs
+    y1: np.ndarray, y2: np.ndarray, axis: None | int | tuple[int] = -1, eps: float = 1e-8, **kwargs
 ) -> np.ndarray:
-    """ Compute the correlation between two NumPy arrays along the specified dimension(s). """
+    """Compute the correlation between two NumPy arrays along the specified dimension(s)."""
```
thomasZen, line 15: I think we should avoid these formatting changes in the future .
Did you do this manually or was this done by black or sth. other script?

openretina/neuron_data_io.py
```diff
@@ -333,7 +310,49 @@ def response_dict(self):
...
+                :,
+            ]
+            validation_mask[
+                (grab_index * self.clip_length) : (grab_index + 1) * self.clip_length,
+                :,
+            ] = False
+
+        if self.use_base_sequence:
+            # Reorder training responses to use the same "base" sequence, which follows the numbering of clips.
+            # This way all training responses are wrt the same order of clips, which can be useful for some applications
```
thomasZen, line 346: Which applications are you thinking of here?

openretina/neuron_data_io.py
```diff
@@ -452,7 +468,120 @@ def _upsample_triggertimes(stim_empirical_duration, stim_theoretical_duration, t
     return upsampled_triggertimes
 
 
-def make_final_responses(data_dict: dict, response_type: Literal["natural", "chirp", "mb"] = "natural"):
+def _apply_mask_to_field(data_dict, field, mask):
```
thomasZen: Maybe readd type annotations

openretina/neuron_data_io.py
```diff
@@ -452,7 +468,120 @@ def _upsample_triggertimes(stim_empirical_duration, stim_theoretical_duration, t
...
+        if key in ["roi_mask", "roi_coords"]:
+            continue
+        if isinstance(data_dict[field][key], np.ndarray) and len(data_dict[field][key]) > 0:
+            if len(data_dict[field][key].shape) == 1:
+                data_dict[field][key] = data_dict[field][key][mask]
+            elif len(data_dict[field][key].shape) == 2:
+                if data_dict[field][key].shape[0] == mask.shape[0]:
+                    data_dict[field][key] = data_dict[field][key][mask, :]
+                else:
+                    data_dict[field][key] = data_dict[field][key][:, mask]
```
thomasZen: Add comments here (e.g. why are there two different cases)

openretina/neuron_data_io.py
```diff
@@ -452,7 +468,120 @@ def _upsample_triggertimes(stim_empirical_duration, stim_theoretical_duration, t
...
+    new_data_dict = deepcopy(data_dict)
+
+    if logic not in ["and", "or"]:
+        raise ValueError("logic must be either 'and' or 'or'")
+    assert len(qi_types) == len(qi_thresholds), "qi_types and qi_thresholds must have the same length"
+
+    for field in new_data_dict.keys():
+        masks = [
+            new_data_dict[field][f"{qi_type}_qi"] >= qi_threshold
+            for qi_type, qi_threshold in zip(qi_types, qi_thresholds)
```
thomasZen:
```suggestion
            for qi_type, qi_threshold in zip(qi_types, qi_thresholds, strict=True)
```
If you use this you probably don't need the assert above.

openretina/neuron_data_io.py
```diff
@@ -452,7 +468,120 @@ def _upsample_triggertimes(stim_empirical_duration, stim_theoretical_duration, t
...
+        ]
+
+        if logic == "and":
+            combined_mask = np.logical_and.reduce(masks)
+        else:  # logic == 'or'
+            combined_mask = np.logical_or.reduce(masks)
```
thomasZen, line 537: Instead of passing "and" or "or" in the arguments you could just path the reduce function, e.g. `np.logical_or.reduce`

openretina/neuron_data_io.py
```diff
@@ -502,4 +655,143 @@ def make_final_responses(data_dict: dict, response_type: Literal["natural", "chi
...
+
+    Returns:
+        Dict[str, dict]: A filtered dictionary of neuron responses that meet the specified criteria.
+    """
+
+    def print_verbose(message):
+        if verbose:
+            print(message)
+
+    def get_n_neurons(all_responses):
```
thomasZen:
```suggestion
    def get_n_neurons(all_responses) -> int:
```

openretina/neuron_data_io.py
```diff
@@ -502,4 +655,143 @@ def make_final_responses(data_dict: dict, response_type: Literal["natural", "chi
...
+            f"(-{(dropped_n_cell_types) / original_neuron_count :.2%})."
+        )
+        print_verbose(" ------------------------------------ ")
+    else:
+        all_rgcs_responses_ct_filtered = deepcopy(all_responses)
+
+    count_before_checks = get_n_neurons(all_rgcs_responses_ct_filtered)
+
+    # Apply quality checks
+    d_qi = d_qi if d_qi is not None else 0.0
```
thomasZen: This sometimes leads to type errors as the type of d_qi will still be `Optional[float]`.
if you run into this issue you can call the argument d_qi_optional and then write sth. like:
`d_qi = d_qi_optional if d_qi _optional is not None else 0.0`

openretina/neuron_data_io.py
```diff
@@ -502,4 +655,143 @@ def make_final_responses(data_dict: dict, response_type: Literal["natural", "chi
...
+    all_rgcs_responses_ct_filtered = _apply_qi_mask(
+        all_rgcs_responses_ct_filtered, ["d", "chirp"], [d_qi, chirp_qi], qi_logic
```
thomasZen, line 737: What's d? maybe give it a more expressive name.

openretina/utils/misc.py
thomasZen: Also we could use pythons logger class, and use logger.debug then.
https://docs.python.org/3/howto/logging.html

Not necessarily for now, but for future uses.

thomasZen (approved):
In principle looks good.
Added questions for some parts, and for the new functions type annotations would be nice

openretina/metrics.py
fededagos: I agree, this is done by Sourcery which I use for refactoring stuff. I manually disabled this refactor now so it should not happen again

openretina/utils/misc.py
fededagos: I used this for the reports I am creating in .HTML. It is handy to disable unnecessary output from jupyter code cells, or when you want to hide a progress bar for example (which is not necessary to show in a rendered report)

openretina/metrics.py
fededagos: It was ruff. I think it is because we changed at some point from max-line-length=80 to 120

openretina/neuron_data_io.py
fededagos: I used this to plot the overlay of the response of all cells in all dataloaders to the same movie stimulus. I was trying it before to also train on the base sequence, but that is 1. A stupid idea and 2. Not working

openretina/neuron_data_io.py
fededagos: Good point, will add.

openretina/neuron_data_io.py
fededagos: Added documentation in the corresponding function, unfortunately it depends on how the qis are saved in the data file

PR #49, opened 2024-08-16
Correlation loss between output traces

thomasZen:
The implementation to compute the correlation should be okay, the loss is still is quite simple.


openretina/correlation_loss.py
```diff
@@ -0,0 +1,29 @@
...
+def calculate_pairwise_correlations(traces: torch.Tensor, unbiased: bool = True) -> torch.Tensor:
+    """ traces.shape = (..., time_steps, neurons) """
+    if unbiased:
+        traces = traces - traces.mean(dim=-2, keepdim=True)
+    norm_traces = torch.nn.functional.normalize(traces, dim=-2)
+    # pairwise_correlations.shape = (..., neurons, neurons)
+    norm_traces_t = norm_traces.transpose(-2, -1)
+    pairwise_correlations = norm_traces_t @ norm_traces
+    # remove diagonal values that are 1.0 
+    one_correlation_mask = pairwise_correlations < 0.999
```
fededagos: This would probably not impact operations later, but in the way I calculated it I did not have repeating elements in there, i.e. I only used the upper or lower triangular parts of the pairwise correlation matrix. In this case, where things are batched, this should work:
```{python}
upper_tri_indices = torch.triu_indices(pairwise_correlations.size(-1), 
                                       pairwise_correlations.size(-1), 
                                       offset=1)
unique_correlations = pairwise_correlations[..., upper_tri_indices[0], upper_tri_indices[1]]
    
return unique_correlations
```
The offset also removes the diagonal directly without having to threshold by a high value

openretina/correlation_loss.py
```diff
@@ -0,0 +1,29 @@
...
+    pairwise_correlations_no_diag = pairwise_correlations[one_correlation_mask]
+    return pairwise_correlations_no_diag
+
+
+def simple_mean_variance_loss(traces: torch.Tensor, unbiased: bool = True) -> torch.Tensor | float:
```
fededagos: I would maybe only add a tunable multiplier hyperparameter to govern how big of an impact this loss term should make on the total loss

openretina/correlation_loss.py
thomasZen: Good point, I switched to above solution.

openretina/correlation_loss.py
thomasZen: I added a interpolation weight in the core readout class.

thomasZen:
Closing as it's outdated.

PR #52, opened 2024-09-10
Separate workflow jobs so they can fail independently of one another

fededagos:


fededagos:
I temporarily removed the status checks branch protection rules until we merge this. Afterwards we can re-enable them, deciding which ones will be necessary (it is Ok with me to re-enable all of them)

.github/workflows/python-tests.yml
```diff
@@ -28,13 +30,19 @@ jobs:
...
         run: |
           ./scripts/tests/codestyle.sh
+
       - name: Type checks
+        if: matrix.task == 'type-checks'
         run: |
           ./scripts/tests/check-types.sh
+
       - name: Units tests
+        if: matrix.task == 'unit-tests'
```
thomasZen, line 45: The downside to this option is that we install the dependencies for each test once, so three times, and this is the step that takes longest.

I tried to use `continue-on-error` in https://github.com/open-retina/open-retina/pull/53, but it seem to not trigger any tests. Did you deactive the tests completely or only deactivate the rule to merge to main only if all tests pass?

thomasZen: Oh, it said invalid workflow file for draft PR with continue-on-error now: https://github.com/open-retina/open-retina/actions/runs/10879563577/workflow

thomasZen (approved):
Looks good, but in case we ever run into Github restrictions (I don't know how many cpu hours we get for free) we can consider using `if always()`.

(Sorry, I had this tab still open but didn't end up submitting the review an hour earlier)

.github/workflows/python-tests.yml
fededagos: I just de-activated the branch protection rule on merge, not the tests themselves. It is a good point that now the dependency job essentially runs three times... Let's stick with this for now and apply `if always()` in case we run into limitations from github.

PR #53, opened 2024-09-16
Testing continue-on-error

thomasZen:


fededagos:
Was this not working already with my original commit? I had already tested it a bit. `continue-on-error` at the top level will mark jobs that fail as successful.

thomasZen:
> Was this not working already with my original commit? I had already tested it a bit. continue-on-error at the top level will mark jobs that fail as successful.

`if always()` seems to to the trick, see: https://stackoverflow.com/questions/62045967/is-there-a-way-to-continue-on-error-while-still-getting-correct-feedback
Should we go for that?

fededagos:
I am not sure. It is true that if you open the checks in detail it now shows which individual ones fail, but here on the PR summary it says "All checks have failed" because it still treats them as one global check. What was wrong with #52?

fededagos:
The benefit of separating them with fail-fast and using a job matrix is that you can also decide which individual steps are required in branch protection, i.e. you could decide that merging is allowed also without codestyle passing.

thomasZen:
Okay, let's merge the original PR (sorry I forgot to submit my review)

thomasZen:
Closing this in favor of https://github.com/open-retina/open-retina/pull/52

PR #55, opened 2024-09-16
Additional evaluation metrics

fededagos:


fededagos:
@thomasZen I am going to add the code to compute the different kind of oracles in here as well and then request for review again.

openretina/eval/metrics.py
```diff
@@ -171,82 +167,65 @@ def MSE_stop3d(model: torch.nn.Module, loader, avg: bool = True, device: str = "
...
-    :param responses: array of shape #cells x time x repetitions
-    :param predictions: array of shape time x #cells
-    :return: array of shape #cells containing oracle scores
+    total_var = np.var(repeated_outputs, axis=(0, 1), ddof=1)
+    repeats_var = np.var(repeated_outputs, axis=1, ddof=1)
+    noise_var = np.mean(repeats_var, axis=0)
+    # Clip. In some bad cases, noise_var can be larger than total_var.
+    explainable_var = np.clip(total_var - noise_var, eps, None)
+    var_ratio = explainable_var / (total_var + eps)
+    return (var_ratio, explainable_var) if return_explainable else var_ratio
```
thomasZen:
```suggestion
    return var_ratio, explainable_var
```

Suggestion: remove `return_explainable` and always return both. That makes the interface of the function cleaner, and the only downside is that you have to write three more characters when you do not want to use the explainable_var:
```
var_ratio, _ = explainable_vs_total_var(repeated_outputs)
```

openretina/eval/oracles.py
```diff
@@ -0,0 +1,82 @@
...
+        loo_oracles.append(oracle_t)
+    oracle = np.stack(loo_oracles)
+
+    oracle_score = correlation_numpy(
+        rearrange(repeated_responses, "t r n -> (t r) n"),
+        rearrange(oracle, "t r n -> (t r) n", t=repeated_responses.shape[0]),
+        axis=0,
+    )
+
+    return (oracle_score, oracle) if return_oracle else oracle_score
```
thomasZen: Same suggestion as above, maybe remove return_oracle boolean argument and always return both the oracle_score and the oracle.

openretina/eval/oracles.py
```diff
@@ -0,0 +1,82 @@
...
+
+    global_mean_response = repeat(responses.mean(axis=2, keepdims=True), "t r _ -> t r n", n=responses.shape[2])
+
+    oracle_mean_corr = correlation_numpy(
+        rearrange(responses, "t r n -> (t r) n"),
+        rearrange(global_mean_response, "t r n -> (t r) n", t=responses.shape[0]),
+        axis=0,
+    )
+
+    return (oracle_mean_corr, global_mean_response) if return_oracle else oracle_mean_corr
```
thomasZen: maybe remove return_oracle

thomasZen (commented):
looks good so far, I already added a suggestions

openretina/data_io/base.py
```diff
@@ -84,6 +85,16 @@ def normalize_train_test_movies(
...
+    Note: The functions casts the input to torch tensors to calculate the mean and standard deviation of large
+    inputs more efficiently.
+    """
+    train_tensor = torch.tensor(train, dtype=torch.float32)
+    test_tensor = torch.tensor(test, dtype=torch.float32)
+    train_mean = train_tensor.mean()
+    train_std = train_tensor.std()
+    train_video_preproc = (train_tensor - train_mean) / train_std
+    test_video = (test_tensor - train_mean) / train_std
+    return train_video_preproc.cpu().numpy(), test_video.cpu().numpy()
```
thomasZen: any reason why to cast to a torch.tensor, but then return numpy?
(.detach() is not needed here?)

thomasZen (approved):
LGTM (only added one minor comment)

openretina/data_io/base.py
fededagos: I missed the call to .detach(), I will add it!
Regarding the cast to tensor, I found that it helped with speed quite a bit, given that the train video array can be pretty large. I suppose even basic operations are dealt with in parallel by default with torch

PR #60, opened 2024-10-14
Use high res video by default

thomasZen:
The high def video resolution is 72x64 with each pixel being 12.5um wide, for a total size of 900x800um.
The previous stimuli was 18x16 which each pixel being 50um wide (same total size).

Previously, the maximal receptive field sizes were:
- First layer: 11*50um = 550um
- Second layer: (11+5)*50um = 800um (in the best case)

Now using maxpooling at every second layer and the kernel sizes `[16, 16, 8, 8]` (the pixel size increases by a factor of 2 after each maxpool operation):
- First layer: 16*12.5 = 200um
- Second layer: 200um + 16 * 25um = 200um + 400um = 600um
- Third layer: 600um + 8 * 25um = 600 + 200um = 800um
- Fourth layer: 800um + 8 * 50um = 800um + 400um = 1200um

## Results with example config
- Training time per epoch: 7:30 min

## Results
When downsampling to 18x16 using `downsample_input_kernel_size: [1, 4, 4]`:
```
Test metric             DataLoader 0             DataLoader 1             DataLoader 2
────────────────────────────────────────────────────────────────────────────────────────────────────────
    test_correlation        0.24354735016822815      0.2808120548725128       0.4720638394355774
        test_loss             838.07666015625         619.1937255859375        81.65926361083984
─────────────────────────────────────────────────────────────────────────────────────────────────────────
```
Without downsampling:
```
       Test metric             DataLoader 0             DataLoader 1             DataLoader 2
─────────────────────────────────────────────────────────────────────────────────────────────────────
    test_correlation        0.24793203175067902      0.29164376854896545      0.4944961369037628
        test_loss            831.5386352539062        621.2250366210938        81.78510284423828
─────────────────────────────────────────────────────────────────────────────────────────────────────
```

## Further changes:
- Set precision to 16 in the Trainer: this speeds up training by a factor of 2
- When running visualization also store a mp4 video of optimized stimulus
- Use padding mode `"same"` to keep output the same size, such that the number of layers does not affect the output shape. To be backwards compatible with the metrics of hoefling 2024, this PR removes the first 30 temporal frames of the core's output.
- Option to downsample the input such that it matches the input size of Hoefling2024.


openretina/models/core_readout.py
```diff
@@ -310,9 +323,10 @@ def __init__(
             spatial_kernel_sizes=tuple(spatial_kernel_sizes),
         )
         # Run one forward path to determine output shape of core
-        core_test_output = self.core.forward(torch.zeros((1, ) + tuple(in_shape)))
+        example_input = torch.zeros((1, ) + tuple(in_shape))
+        core_test_output = self.core.to(device).forward(example_input.to(device))
         in_shape_readout: tuple[int, int, int, int] = core_test_output.shape[1:]  # type: ignore
-        print(f"{in_shape_readout=}")
+        print(f"{example_input.shape[1:]=} {in_shape_readout=}")
```
thomasZen, line 346: This used to run a forward path on CPU before, and took quite a bit of time with the high res input.

openretina/models/core_readout.py
```diff
@@ -100,6 +100,9 @@ def get_mask(self) -> torch.Tensor:
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         masks = self.get_mask().permute(1, 2, 0)
         y = torch.einsum("nctwh,whd->nctd", x, masks)
+
+        if self.positive:
+            self.features.data.clamp_(0)
```
thomasZen, line 105: This was a bug in the previous code

openretina/models/core_readout.py
```diff
@@ -388,5 +411,5 @@ def configure_optimizers(self):
         }
 
     def save_weight_visualizations(self, folder_path: str) -> None:
-        self.core.save_weight_visualizations(os.path.join(folder_path, "core"))
-        self.readout.save_weight_visualizations(os.path.join(folder_path, "readout"))
+        self.core.save_weight_visualizations(os.path.join(folder_path, "weights_core"))
+        self.readout.save_weight_visualizations(os.path.join(folder_path, "weights_readout"))
```
thomasZen, line 423: I did this change because the plots of the weights will then be names similar and appear next to each others when sorted by folder name

requirements.txt
```diff
@@ -14,6 +14,7 @@ lightning
 wandb
 hydra-core
 h5py
+opencv-python
```
thomasZen, line 17: Dependency for writing mp4 videos.

scripts/train_core_readout.py
```diff
@@ -19,8 +20,9 @@
 
 @hydra.main(version_base=None, config_path="../example_configs", config_name="train_core_readout")
 def main(conf: DictConfig) -> None:
+    torch.set_float32_matmul_precision('medium')
```
thomasZen, line 23: Explicilty setting this got rid of a warning on A100 GPUs.

scripts/train_core_readout.py
```diff
@@ -49,15 +50,18 @@ def main(conf: DictConfig) -> None:
...
-    callbacks = []
+    model_checkpoint = ModelCheckpoint(monitor="val_correlation", save_top_k=3, mode="max", verbose=True)
+    callbacks: list = [model_checkpoint]
     for callback_params in conf.get("training_callbacks", {}).values():
         callbacks.append(hydra.utils.instantiate(callback_params))
 
-    trainer = lightning.Trainer(max_epochs=conf.max_epochs, default_root_dir=conf.save_folder,
-                                logger=logger_array, callbacks=callbacks)
+    trainer = lightning.Trainer(max_epochs=conf.max_epochs, default_root_dir=conf.save_folder, precision=16,
+                                logger=logger_array, callbacks=callbacks, accumulate_grad_batches=1)
```
thomasZen, line 59: precision 16 instead of 32 gives roughly a 2x speedup.

scripts/train_core_readout.py
```diff
@@ -49,15 +50,18 @@ def main(conf: DictConfig) -> None:
...
     trainer.fit(model=model, train_dataloaders=train_loader,
                 val_dataloaders=valid_loader)
-    trainer.test(model, dataloaders=[train_loader, valid_loader, test_loader])
+    # test
+    test_loader = torch.utils.data.DataLoader(LongCycler(dataloaders["test"], shuffle=False), **conf.dataloader)
+    trainer.test(model, dataloaders=[train_loader, valid_loader, test_loader], ckpt_path="best")
```
thomasZen, line 64: I moved the test loader to the end, as each dataloader can use multiple processes, and I don't want that these processes potentially get created before they are needed.

openretina/hoefling_2024/models.py
```diff
@@ -141,7 +141,7 @@ def __init__(
 
         self.features = nn.Sequential()
         if stack is None:
-            self.stack = range(self.layers)
+            self.stack = list(range(self.layers))
         else:
             self.stack = [range(self.layers)[stack]] if isinstance(stack, int) else stack
```
thomasZen, line 146: This lead to failing type checks before, as above the type was range int, but in the else condition it's a list

configs/example_train_core_readout.yaml
fededagos, line 1: Can you make this maybe a separate config e.g. with the suffix `large_stim`or similar? I think it is still valuable to keep the old model for quick access in lightning.

openretina/hoefling_2024/configs.py
fededagos, line 1: Wouldn't removing the norm constraint entirely mess with previous experiments? e.g. the reproduction of Lara's paper? @lhoefling can you comment on this constraint?

scripts/train_core_readout.py
fededagos, line 1: Like with the config, I would also maybe make this separate and call it something like "high_res", or, alternatively, we could make the config composable to distinguish between the two cases but still keep them both saved in the repo somehow. I agree about going to high res as a default, but it is also good to have everything to quickly replicate Lara's model at hand.

scripts/visualize_model_neurons.py
```diff
@@ -78,7 +78,7 @@ def main(
     model.eval()
 
     # from controversial stimuli: (2, 50, 18, 16): (channels, time, height, width)
-    stimulus_shape = (1, 2, 50, 18, 16)
+    stimulus_shape = (1, 2, 50, 72, 64)
```
fededagos: Making this a param of the script would be nice

thomasZen: Good point, I did that.

scripts/train_core_readout.py
thomasZen: I made put the file paths to the movies and the responses in the config.
That should be everything that is different between the high res and low res case in this file.

openretina/hoefling_2024/configs.py
thomasZen: the norm constraints are only used during stimulus optimization.
We had this config duplicated both in the constants and config module, and the one I changed was not used. I deleted this dict from the constants module and consistently use it from the config module now. I also put norm 30 again.

I might tune this again in the future, though. When using the high res stimulus the norm is also higher (my calculation is that the desired norm should be 4 times higher as the norm calculated as`sqrt(sum(stimulus))`,  and the stimulus has 4x4 more values now, but in reality it only seemed to be 1.5 times more at initialization).

fededagos (approved):

PR #61, opened 2024-10-24
Added GRU core to lightning models

fededagos:


configs/model/base_gru_model_lightning.yaml
```diff
@@ -0,0 +1,28 @@
...
+readout_bias: true
+readout_gaussian_masks: true
+readout_gaussian_mean_scale: 6.0
+readout_gaussian_var_scale: 4.0
+readout_positive: true
+readout_gamma: 0.4
+readout_gamma_masks: 0.1
+readout_reg_avg: false
+learning_rate: 0.01
+core_gru_kwargs: null
```
thomasZen: Maybe add new line at the end of the file

openretina/hoefling_2024/models.py
```diff
@@ -514,12 +520,13 @@ def save_weight_visualizations(self, folder_path: str) -> None:
...
-            fig_axes_tuple = plt.subplots(ncols=2, figsize=(2*6, 6))
+            fig_axes_tuple = plt.subplots(ncols=2, figsize=(2 * 6, 6))
             axes: list[plt.Axes] = fig_axes_tuple[1]  # type: ignore
 
             axes[0].set_title("Readout Mask")
-            axes[0].imshow(mask_neuron, interpolation="none", cmap="RdBu_r",
-                           norm=Normalize(-mask_abs_max, mask_abs_max))
+            axes[0].imshow(
+                mask_neuron, interpolation="none", cmap="RdBu_r", norm=Normalize(-mask_abs_max, mask_abs_max)
+            )
```
thomasZen, line 529: Are these changes all done by black (then it doesn't make so much sense to review them I guess)

openretina/models/core_readout.py
```diff
@@ -97,8 +98,12 @@ def get_mask(self) -> torch.Tensor:
         pdf = torch.nan_to_num(pdf / normalisation)
         return pdf
 
+    @property
+    def masks(self) -> torch.Tensor:
+        return self.get_mask().permute(1, 2, 0)
```
thomasZen, line 103: I don't like that we have the public functions masks and get_masks() now. I think we should only have one of them.

openretina/models/core_readout.py
```diff
@@ -207,32 +213,36 @@ def save_weight_visualizations(self, folder_path: str) -> None:
...
 
         super().__init__()
         self.gamma_input = gamma_input
         self.gamma_temporal = gamma_temporal
         self.gamma_in_sparse = gamma_in_sparse
 
         self.features = torch.nn.Sequential()
-        for layer_id, (num_in_channels, num_out_channels) in enumerate(
-                zip(channels[:-1], channels[1:], strict=True)):
+        for layer_id, (num_in_channels, num_out_channels) in enumerate(zip(channels[:-1], channels[1:], strict=True)):
```
thomasZen: Are these changes necessary, they probably lead to merge conflicts with https://github.com/open-retina/open-retina/pull/60/

openretina/models/gru_core.py
```diff
@@ -0,0 +1,567 @@
...
+    STSeparableBatchConv3d,
+    TimeIndependentConv3D,
+    TimeLaplaceL23dnorm,
+    TorchFullConv3D,
+    TorchSTSeparableConv3D,
+    compute_temporal_kernel,
+    temporal_smoothing,
+)
+
+DEVICE = "cuda:0" if torch.cuda.is_available() else "cpu"
```
thomasZen: Do you actually use this, I think it should be mostly managed by pytorch ligthning now?
Maybe delete

thomasZen (commented):
Small comments so far. We should also decide whether to use black or not, if we decide to do so then everyone has to use it. Otherwise, we run into annoying merge conflicts for no good reason.
I'll read through the gru_core.py file after lunch.

openretina/models/gru_core.py
```diff
@@ -0,0 +1,567 @@
...
+        input_kern,
+        rec_kern,
+        groups=1,
+        gamma_rec=0,
+        pad_input=True,
+        **kwargs,
+    ):
+        super().__init__()
+
+        input_padding = input_kern // 2 if pad_input else 0
```
thomasZen, line 64: Only as a note: this is not necessary the same as "same" padding in pytorch, as for for odd (or even) inputs it gives you an output that is off by 1 from the input.

openretina/models/gru_core.py
```diff
@@ -0,0 +1,567 @@
...
+
+        # generate empty prev_state, if None is provided
+        if prev_state is None:
+            prev_state = self.init_state(input_)
+
+        update = self.update_gate_input(input_) + self.update_gate_hidden(prev_state)
+        update = F.sigmoid(update)
+
+        reset = self.reset_gate_input(input_) + self.reset_gate_hidden(prev_state)
+        reset = F.sigmoid(reset)
```
thomasZen:
```suggestion
        update_gate = self.update_gate_input(input_) + self.update_gate_hidden(prev_state)
        update_gate = F.sigmoid(update_gate)

        reset_gate = self.reset_gate_input(input_) + self.reset_gate_hidden(prev_state)
        reset_gate = F.sigmoid(reset_gate)
```

This makes it more clear that it's only the gate, but not the content that is written to the output.

openretina/models/gru_core.py
```diff
@@ -0,0 +1,567 @@
...
+            x = feat(
+                (
+                    x,
+                    data_key,
+                )
+            )
+            if hasattr(feat, "conv") and hasattr(feat, "norm"):
+                x = getattr(self, f"film_{layer_num}")(x, conditioning)
+
+        return
```
thomasZen:
```suggestion
        return x
```

This function seems like it should return something!

thomasZen (approved):
LGTM modulo missing return value

openretina/hoefling_2024/models.py
fededagos: Not black but indeed Ruff formatting. I have it enabled by default on save :)

openretina/models/core_readout.py
fededagos: I have implemented it for convenience (I have used it in the session agnostic model) and for consistency with the old models, which always had a "masks" attribute. I agree that it feels a bit redundant, but maybe `get_masks` should be called `compute_gaussian_masks`, and then both could be kept. Not sure what is the best way.

openretina/models/gru_core.py
fededagos: Great point, it was copied over from dev_models...

openretina/models/gru_core.py
fededagos: Another copying error...

openretina/models/core_readout.py
fededagos: These are also changes applied by ruff upon saving. Should I revert them?

PR #65, opened 2024-11-07
Dev: Test ensemble model loading

thomasZen:
This test just checks if the model can be downloaded from the GIN node and then runs a forward path using the downloaded model.

This is useful as this test will break if somebody renames the classes the model uses, e.g. renaming `EnsembleModel` to `EnsembleModels` in the following code will make the test fail:
https://github.com/open-retina/open-retina/blob/e143c8720ba261b1c9ee6e3e5fa0fa8c019aec25/openretina/hoefling_2024/nnfabrik_model_loading.py#L295


fededagos (approved):
Looks good. For the new lightning models we could do the same but also checking with simplified real data

PR #66, opened 2024-11-11
Reset adam parameters on LR reset, better logging and MovieSampler fix

thomasZen:
- Re-initialize the Adam if the learning rate was changed by using a new class called OptimizerResetCallback
- Refactor MovieSampler and generate shifts from 0 to clip_chunk_size when not enforcing scene boundaries
- Better logging:
   - Log learning rate
   - Log training stats per epoch leading to smoother and easier comparable graphs
   - Log gradient magnitudes
- Change precision=16 to precision="16-mixed" as this is recommended by pytorch lightning
- Filter data by default based by removing low quality cells

# Results

## Low res
```
───────────────────────────────────────────────────────────────────────────────────────────────────────────────
       Test metric             DataLoader 0             DataLoader 1             DataLoader 2                                                                                                               
───────────────────────────────────────────────────────────────────────────────────────────────────────────────
    test_correlation        0.22427836060523987      0.2659996449947357       0.4651343822479248                                                                                                            
        test_loss            534.9968872070312        375.6442565917969        38.34758377075195                                                                                                            
───────────────────────────────────────────────────────────────────────────────────────────────────────────────
```
With seed 42:
```
    test_correlation        0.20960338413715363      0.24668897688388824      0.4416184425354004                                                                               
        test_loss            535.2044067382812        377.0640869140625        37.99009704589844 
```

## High Res
With seed 42 (but one layer does not support deterministic mode):
```
────────────────────────────
       Test metric             DataLoader 0             DataLoader 1             DataLoader 2
───────────────────────────────────────────────────
    test_correlation        0.28346744179725647      0.33637067675590515       0.524970531463623
        test_loss            520.753662109375         364.3420715332031        38.09238052368164
────────────────────────────────────────────────────────────
```

### DIFF
--- a/configs/example_train_core_readout_low_res.yaml
+++ b/configs/example_train_core_readout_low_res.yaml
@@ -20,7 +20,7 @@ core_readout:
   gaussian_masks: True
   gaussian_mean_scale: 6.0
   gaussian_var_scale: 4.0
-  positive: False
+  positive: True
   gamma_readout: 0.4
   learning_rate: 0.01
   cut_first_n_frames_in_core: 30
--- a/configs/example_train_core_readout_low_res.yaml
+++ b/configs/example_train_core_readout_low_res.yaml
@@ -8,6 +8,7 @@ movies_filename: "2024-01-11_movies_dict_8c18928.pkl"
 responses_filename: "2024-08-14_neuron_data_responses_484c12d_djimaging.h5"
 save_folder: exp/
 max_epochs: 100
+seed: 42
 
 core_readout:
   in_channels: 2
--- a/configs/example_train_core_readout_low_res.yaml
+++ b/configs/example_train_core_readout_low_res.yaml
@@ -1,4 +1,8 @@
-exp_name: example_core_readout_low_res
+defaults:
+  - _self_
+  - quality_checks: hoefling_2024
+
+exp_name: example_core_readout_low_res_filtered
 data_folder: "/gpfs01/euler/data/SharedFiles/projects/TP12/"
 movies_filename: "2024-01-11_movies_dict_8c18928.pkl"
 responses_filename: "2024-08-14_neuron_data_responses_484c12d_djimaging.h5"
@@ -16,7 +20,7 @@ core_readout:
   gaussian_masks: True
   gaussian_mean_scale: 6.0
   gaussian_var_scale: 4.0
-  positive: True
+  positive: False
   gamma_readout: 0.4
   learning_rate: 0.01
   cut_first_n_frames_in_core: 30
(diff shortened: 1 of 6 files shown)


configs/example_train_core_readout_low_res.yaml
```diff
@@ -16,7 +20,7 @@ core_readout:
   gaussian_masks: True
   gaussian_mean_scale: 6.0
   gaussian_var_scale: 4.0
-  positive: True
+  positive: False
```
fededagos: Why? Lara's original config seems to be positive: https://github.com/open-retina/open-retina/blob/c2e60e2bf5189fd8a2491d01247621cc456be20b/openretina/hoefling_2024/configs.py#L8

openretina/models/core_readout.py
```diff
@@ -373,6 +374,20 @@ def __init__(
...
 
+    def on_before_optimizer_step(self, optimizer):
+        # Compute the 2-norm for each layer
+        # If using mixed precision, the gradients are already unscaled here
+        core_norms = grad_norm(self.core, norm_type=2)
+        self.log_dict(core_norms, on_step=False, on_epoch=True)
+        readout_norms = grad_norm(self.readout, norm_type=2)
+        self.log_dict(readout_norms, on_step=False, on_epoch=True)
+
+    def get_last_lr(self) -> float:
```
fededagos: There is an already implemented callback for lightning to do this: https://pytorch-lightning.readthedocs.io/en/1.2.10/api/pytorch_lightning.callbacks.lr_monitor.html Could we use it? Would avoid having to add this manually in the model.

It should also be more robust given that it is a lightning callback to use out of the box. You could maybe add this to the example config directly.

I think you could just copy these lines from one of the configs in the cell-type branch and it should work:
https://github.com/open-retina/open-retina/blob/c4d3f8527af77358055b399aa5ab9b4a0322d083/configs/minimal_session_agnostic_lightning.yaml#L52

fededagos (changes requested):
Could you add the lr_monitor as a callback? Everything else looks good, though I also do not know why you changed readout_positive: False in Lara's config.

configs/example_train_core_readout_low_res.yaml
thomasZen: Good point, I changed this back.

openretina/models/core_readout.py
thomasZen: Nice, I changed the config and got rid of this function

thomasZen:
I also added the option to use a seed and then set the determistic option in pytorch lightning

fededagos (approved):
All good :)

PR #67, opened 2024-11-13
Support for data from Maheswaranathan 2023 paper

fededagos:
- Refactored previous code regarding Maheswaranathan data
- Added a simpler custom dataloader for it
- Custom config and lightning training script

Training results (no tweaking):
| Test metric           | DataLoader 0         | DataLoader 1         | DataLoader 2         |
|-----------------------|----------------------|----------------------|----------------------|
| test_correlation      | 0.17187398672103882  | 0.30929845571517944  | 0.6416633725166321   |
| test_loss             | -766.4127807617188   | -592.941162109375    | -60.24884033203125   |

(Performance in the paper with hyperparameter tuning is 0.7, so not too bad)


openretina/cyclers.py
fededagos, line 1: This is to use multiple workers on the cycler without duplicating the underlying data.

openretina/hoefling_2024/data_io.py
```diff
@@ -17,37 +17,19 @@ class MoviesDict(TypedDict):
     random_sequences: Optional[np.ndarray]
 
 
-def get_all_movie_combinations(
+def generate_movie_splits(
```
fededagos, line 20: I have split the function into multiple ones while keeping the functionality. This way can reuse the part to generate the splits elsewhere

openretina/cyclers.py
```diff
@@ -35,16 +36,29 @@ def __init__(self, loaders: Dict[str, DataLoader], shuffle: bool = True):
...
+            num_workers = worker_info.num_workers
+            worker_id = worker_info.id
+            total_iterations = len(self.loaders) * self.max_batches
+            per_worker = (total_iterations + num_workers - 1) // num_workers
+            iter_start = worker_id * per_worker
+            iter_end = min(iter_start + per_worker, total_iterations)
+
+        # Yield batches in the assigned range
+        for k, loader, _ in islice(zip(cycle(keys), cycle(cycles), range(total_iterations)), iter_start, iter_end):
             yield k, next(loader)
```
thomasZen, line 62: Was the previous implementation of this somewhat faulty?
I.e. one epoch iterated once through the data loader but previously iterated num worker times through the data?

openretina/dataloaders.py
```diff
@@ -157,11 +233,16 @@ def get_movie_dataloader(
         print("Nans in responses, skipping this dataloader")
         return None
 
-    if scene_length is not None and split == "train" and chunk_size > scene_length:
+    if scene_length is not None and split == "train" and chunk_size > scene_length and not allow_over_boundaries:
```
thomasZen:
```suggestion
    if not allow_over_boundaries and scene_length is not None and split == "train" and chunk_size > scene_length:
```
I think not allow over boundaries is the context to understand the next to conditions, so putting it first makes it easier to read this line.

openretina/dataloaders.py
```diff
@@ -192,6 +274,50 @@ def get_movie_dataloader(
...
+    elif split == "test":
+        if isinstance(movies, dict):
+            start_indices = {k: [0] for k in movies.keys()}
+        else:
+            start_indices = [0]
+
+    else:
+        raise NotImplementedError("Start indices could not be recovered.")
+
+    return start_indices
```
thomasZen: I think it would be cleaner to split this function into two functions, one for the dictionary case and one for the non dictionary case.
This would also make the type annotations way easier and will get rid of the isinstance calls in this function.

openretina/hoefling_2024/data_io.py
```diff
@@ -71,47 +52,88 @@ def get_all_movie_combinations(
...
+):
+    movies = {
+        "left": {
+            "train": {},
+            "validation": torch.flip(movie_val, [-1]),
+            "test": torch.flip(movie_test, [-1]),
+        },
+        "right": {"train": {}, "validation": movie_val, "test": movie_test},
+        "val_clip_idx": val_clip_idx,
+    }
```
thomasZen, line 76: not necessarily for now, but in the future we could think of defining a class structure like:
```
class RetinaMovie:
left: DataSplit
right: DataSplits
val_clip_idx: int

class DataSplit:
   train: ...
   test: ...
   validation: ...
```
This will make it easier to reason about the data structures in downstream functions.

openretina/hoefling_2024/data_io.py
```diff
@@ -71,47 +52,88 @@ def get_all_movie_combinations(
...
+# Main function that assembles everything
+def get_all_movie_combinations(
+    movie_train,
+    movie_test,
+    random_sequences: np.ndarray,
+    val_clip_idx: Optional[List[int]] = None,
+    num_clips: int = NUM_CLIPS,
+    num_val_clips: int = NUM_VAL_CLIPS,
+    clip_length: int = CLIP_LENGTH,
+    seed=1000,
```
thomasZen:
```suggestion
    seed: int =1000,
```

openretina/maheswaranathan_2023/data_io.py
```diff
@@ -0,0 +1,110 @@
+from typing import Any, List, Optional
+
+import numpy as np
+from tqdm.auto import tqdm
+
+from openretina.dataloaders import get_movie_dataloader
+from openretina.hoefling_2024.data_io import MoviesDict, generate_movie_splits
+from openretina.maheswaranathan_2023.neuron_data_io import NeuronDataBaccus
+
+from .constants import CLIP_LENGTH
```
thomasZen:
```suggestion
from openretina.maheswaranathan_2023.constants import CLIP_LENGTH
```
I'm a bit more in favor of absolute imports, especially since we have multiple constants files in our repo

openretina/maheswaranathan_2023/data_io.py
```diff
@@ -0,0 +1,110 @@
...
+    ), "neuron_data_dictionary should be a dictionary of sessions and their corresponding neuron data."
+
+    assert (
+        neuron_data_dictionary.keys() == movies_dictionary.keys()
+    ), "The keys of neuron_data_dictionary and movies_dictionary should match."
+
+    assert all(
+        "train" in movies_dictionary[session_key] and "test" in movies_dictionary[session_key]
+        for session_key in movies_dictionary
+    ), "movies_dictionary should contain train and test movies for all sessions."
```
thomasZen, line 55: Coupling these in a class structure could enforce that easier and at a single position (e.g. in a isValid method in the class or on initalization)

openretina/maheswaranathan_2023/data_io.py
```diff
@@ -0,0 +1,110 @@
...
+
+    assert all(field in next(iter(neuron_data_dictionary.values())) for field in ["responses_final", "stim_id"]), (
+        "Check the neuron data dictionary sub-dictionaries for the minimal"
+        " required fields: 'responses_final' and 'stim_id'."
+    )
+
+    #! Todo num_clips needs to be determined on the dataset.
```
thomasZen: todo

openretina/maheswaranathan_2023/data_io.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Initialise dataloaders
+    dataloaders: dict[str, Any] = {"train": {}, "validation": {}, "test": {}}
+
+    for session_key, session_data in tqdm(neuron_data_dictionary.items(), desc="Creating movie dataloaders"):
+        # Extract all data related to the movies first
+        num_clips = movies_dictionary[session_key]["train"].shape[1] // clip_length
+
+        # Draw validation clips based on the random seed
+        rnd = np.random.RandomState(seed)
```
thomasZen, line 70: maybe we should rely on lightning seed_everything method instead of doing it ourselfes: https://pytorch-lightning.readthedocs.io/en/1.7.7/api/pytorch_lightning.utilities.seed.html#pytorch_lightning.utilities.seed.seed_everything

Would also get rid of passing the seed through the whole hierarchy.

openretina/maheswaranathan_2023/neuron_data_io.py
```diff
@@ -0,0 +1,154 @@
...
+"""
+
+import os
+from typing import Any, List, Optional
+
+import numpy as np
+import torch
+from jaxtyping import Float
+
+from ..utils.h5_handling import load_dataset_from_h5
```
thomasZen:
```suggestion
from openretina.utils.h5_handling import load_dataset_from_h5
```

openretina/maheswaranathan_2023/neuron_data_io.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+
+        self.key = key
+        self.roi_coords = ()
+        self.clip_length = clip_length
+        self.num_clips = num_clips
+        self.val_clip_idx = val_clip_idx
+
+    # this has to become a regular method in the future!
+    @property
```
thomasZen: make it one now, with type checks it should be relatively fast to verify that you changed all calls.

openretina/maheswaranathan_2023/neuron_data_io.py
```diff
@@ -0,0 +1,154 @@
...
+    def response_dict(self):
+        # Transpose the responses to have the shape (n_timepoints, n_neurons)
+        self.responses_test = self.neural_responses["test"].T
+        self.responses_train_and_val = self.neural_responses["train"].T
+        self.test_responses_by_trial = []
+
+        self.compute_validation_responses()
```
thomasZen: Pass arguments directly to the function instead of setting the class variables.
It's much easier to see what's going on if the function takes the arguments and returns the results instead of doing it implicitly via the state of the class...

openretina/maheswaranathan_2023/neuron_data_io.py
```diff
@@ -0,0 +1,154 @@
...
+        return {
+            "train": torch.tensor(self.responses_train).to(torch.float),
+            "validation": torch.tensor(self.responses_val).to(torch.float),
+            "test": {
+                "avg": torch.tensor(self.responses_test).to(torch.float),
+                "by_trial": torch.tensor(self.test_responses_by_trial),
+            },
+        }
+
+    def compute_validation_responses(self) -> None:
```
thomasZen: this function should return something

openretina/maheswaranathan_2023/neuron_data_io.py
```diff
@@ -0,0 +1,154 @@
...
+                    "test": test_video,
+                }
+
+                train_session_data = load_dataset_from_h5(recording_file, f"/train/response/{response_type}")
+                test_session_data = load_dataset_from_h5(recording_file, f"/test/response/{response_type}")
+
+                responses_all_sessions["".join(session.split("/")[-1])] = {
+                    "responses_final": {
+                        "train": train_session_data[:, : train_video.shape[1]] / fr_normalization,
+                        "test": test_session_data[:, : test_video.shape[1]] / fr_normalization,
```
thomasZen: maybe double check why we need the `:train_video.shape[1]`

thomasZen (approved):
Looks mostly reasonable.
I feel strongly about using arguments and return values for functions instead of using the object state (i.e in compute_validation_responses), apart from that decide freely if you want to add the suggestion :)

openretina/cyclers.py
fededagos: I believe the biggest issue with not adapting the implementation (i.e. before) was that each worker would have a copy of the entire iterators (i.e. all sessions), multiplying the memory requirements. I am not sure if this also meant that the model also saw duplicate data... Maybe yes. See https://discuss.pytorch.org/t/iterable-pytorch-dataset-with-multiple-workers/135475/2

openretina/maheswaranathan_2023/neuron_data_io.py
fededagos: I agree, this is ugly. I will change it. I believe the same is happening in the standard code for Lara's model. I will have a look there as well and remove.

openretina/maheswaranathan_2023/data_io.py
fededagos: I really like this, but it needs to be added at the beginning of each script I suppose? What is the best way to do this? Starting an issue...

openretina/maheswaranathan_2023/neuron_data_io.py
fededagos: I actually like this as a property in the end, but I refactored the rest

PR #71, opened 2024-11-18
Refactoring PR

thomasZen:
See Readme for new structure. I put everything for which I'm unsure if we need it for the paper in openretina/legacy or scripts/legacy (I will also go over these files again, though. Sometimes I moved them to legacy because I did not have time to make them work.). I think having a rather small but clean repository is best, so feel free to suggest if we should move more to legacy but also comment what we should keep. My suggestion is: we can leave it there and re-evaluate if we should throw these away or integrate them into the new repo structure again.

Further changes:
- Put mypy config in pyproject.toml and remove scripts/tests
- Do not manually set seed in dataloaders. This is managed by pytorch ligthning/seed everything, now.
- Explicitly specify readout input dimension in config


openretina/data_io/cyclers.py
```diff
@@ -48,6 +47,7 @@ def __iter__(self):
         if worker_info is None:  # Single-process data loading
             iter_start = 0
             iter_end = len(self.loaders) * self.max_batches
+            total_iterations = iter_end
```
thomasZen, line 50: This was a bug before.

configs/example_train_core_readout.yaml
```diff
@@ -13,9 +13,9 @@ seed: 42
 core_readout:
   in_channels: 2
   features_core: [16, 16, 16, 16]
-  temporal_kernel_sizes: [16, 16, 8, 8]
-  spatial_kernel_sizes: [16, 16, 8, 8]
-  in_shape: [2, 150, 72, 64]
+  temporal_kernel_sizes: [15, 15, 7, 7]
+  spatial_kernel_sizes: [15, 15, 7, 7]
```
thomasZen, line 17: Pytorch lightning says that odd kernel sizes are more efficient when using padding='same'

configs/maheswaranathan_core_readout.yaml
```diff
@@ -2,6 +2,7 @@ defaults:
   - model: base_gru_model_lightning
 
 exp_name: core_readout_maheswaranathan
+seed: 42
```
thomasZen, line 5: We call seed_everything with this seed then (Maybe there's a even better option, but this one seemed to set seeds also on worker threads that do dataloading)

fededagos (approved):

PR #45, opened 2024-08-15
Lightning interface for the core readout model

thomasZen:
- Simplify Readout layer
- Add lightning interface to the core readout model
- Use config for training

## Training on the cluster
Run the following command from any directory:
```
~/GitRepos/open-retina/scripts/train_core_readout.py \
--config-path ~/GitRepos/open-retina/configs \
--config-name example_train_core_readout
```
The current scores with this approach are (0=train, 1=validation, 2=test):
```
────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       Test metric             DataLoader 0             DataLoader 1             DataLoader 2                                                                                                               
────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    test_correlation        0.27609729766845703      0.3052205443382263       0.5100897550582886                                                                                                            
        test_loss            817.4778442382812        627.6156005859375        82.57406616210938                                                                                                            
────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

## Config
I used [hydra](https://hydra.cc/docs/intro/) for config parsing. This has a few advantages:
- You can dynamically instantiate objects, see loggers in example_config
- You can overwrite any config parameter from the command line, see [hydra docs](https://hydra.cc/docs/advanced/override_grammar/basic/)

## Todo
- Spatial Laplace is not implemented

PR #72, opened 2024-11-21
Fix cyclers and move validation_clip_indices to config

thomasZen:
I think the way we used and implemented the cyclers had a few downside.

For the short cycler I added multiple worker support. For the long cycler I think it is not so important as it's okay to see examples more often during training as long as it's randomly sampled. I added a test for both the long and the short cycler that will make it easy to verify if we want to add this feature.

I also define the validation clip indices in the config now, that made it possible to remove the num validation clip constant. The validation clip indices stated in the config were exactly the same we randomly generated before with the seed 42. This also allows to change the validation clip indices easily in the config (e.g. to test if using less validation clips and therefore increasing the training data helps for the performance on the test set).

## We can duplicate data during validation and testing
In the long cycler we use examples from smaller session more often and iterate exactly once over examples from smaller sessions.

## Bug when using multiple workers in LongCycler
The following code is sketchy as it **only** calls next inside the loop:
https://github.com/open-retina/open-retina/blob/bd7ebb567d2dc7b7b41b5a599701850466e1c953/openretina/data_io/cyclers.py#L61-L62
When using multiple workers, this means we only look at the first example(s) of each dataloader.
This is a bug, but I do not know if it has an effect as we also randomize the examples of the individual dataloaders during training.


configs/example_train_core_readout.yaml
```diff
@@ -29,6 +29,10 @@ core_readout:
   dropout_rate: 0.0
   # downsample_input_kernel_size: [1, 4, 4]  # to reproduce Hoefling input size of 18x16px
 
+natmov_dataloader:
+    validation_clip_indices: [0, 4, 10, 11, 18, 30, 45, 62, 67, 77, 79, 80, 81, 83, 95]
+    train_chunk_size: 100
```
fededagos, line 34: I am sure this was 50 before. This alone might explain why now train is doing better and val is not higher than train anymore like we used to see.

configs/example_train_core_readout_low_res.yaml
```diff
@@ -27,6 +27,10 @@ core_readout:
   cut_first_n_frames_in_core: 30
   dropout_rate: 0.0
 
+natmov_dataloader:
+    validation_clip_indices: [0, 4, 10, 11, 18, 30, 45, 62, 67, 77, 79, 80, 81, 83, 95]
+    train_chunk_size: 100
```
fededagos, line 32: See comment before

fededagos (approved):
All good!

configs/example_train_core_readout.yaml
thomasZen: For pytorch lightning this was 100 from the start (somehow): https://github.com/open-retina/open-retina/pull/45/files#r1856341784

I'll leave it at this value for now, but we can change it further in case it makes a difference.

PR #73, opened 2024-11-22
Refactored base classes for data_io

### LINKED ISSUE #68: More readable annotations for movie stimuli
not necessarily for now, but in the future we could think of defining a class structure like:
```
class RetinaMovie:
left: DataSplit
right: DataSplits
val_clip_idx: int

class DataSplit:
   train: ...
   test: ...
   validation: ...
```
This will make it easier to reason about the data structures in downstream functions.

_Originally posted by @thomasZen in https://github.com/open-retina/open-retina/pull/67#discussion_r1841897865_

fededagos:
Hopefully more intuitive way of dealing with data_io input classes. It essentially assumes that responses fed to "dataloader-making" functions are in a dictionary, where keys are the session names and values are "ResponsesTrainTestSplit" objects, holding train and test responses for each session.

This was less straightforward than expected as for Lara's data the test responses were computed during dataloader creation, and not before.

fixes #68


openretina/data_io/base.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+
+    def check_matching_stimulus(self, movies: MoviesTrainTestSplit):
+        assert self.train.shape[1] == movies.train.shape[1], "Train movie and responses should have the same timebins."
+        assert self.test.shape[1] == movies.test.shape[1], "Test movie and responses should have the same timebins."
+        if self.stim_id is not None and movies.stim_id is not None:
+            assert self.stim_id == movies.stim_id, "Stimulus ID in responses and movies do not match."
+
+    @property
+    def n_neurons(self):
```
thomasZen:
```suggestion
    def n_neurons(self) -> int:
```

openretina/data_io/base.py
```diff
@@ -0,0 +1,69 @@
...
+
+import numpy as np
+from jaxtyping import Float
+
+
+@dataclass
+class MoviesTrainTestSplit:
+    train: Float[np.ndarray, "channels train_time height width"]
+    test: Float[np.ndarray, "channels test_time height width"]
+    stim_id: Optional[str] = None
```
thomasZen, line 14: Is this ever used? If not remove it.

openretina/data_io/base_dataloader.py
```diff
@@ -369,3 +352,87 @@ def get_movie_dataloader(
...
+            validation_mask[
+                (grab_index * self.clip_length) : (grab_index + 1) * self.clip_length,
+                :,
+            ] = False
+
+        responses_train = self.responses_train_and_val[validation_mask].reshape(-1, self.num_neurons)
+        return responses_train, responses_val
+
+    @property
+    def response_dict(self):
```
thomasZen:
```suggestion
    def response_dict(self) -> dict:
```

openretina/data_io/base_dataloader.py
```diff
@@ -369,3 +352,87 @@ def get_movie_dataloader(
...
+        self.val_clip_idx = val_clip_idx
+
+        # Transpose the responses to have the shape (n_timepoints, n_neurons)
+        self.responses_test = self.neural_responses.test.T
+        self.responses_train_and_val = self.neural_responses.train.T
+
+        self.responses_train, self.responses_val = self.compute_validation_responses()
+        self.test_responses_by_trial = np.array([])  # Added for compatibility with Hoefling et al., 2024
+
+    def compute_validation_responses(self) -> tuple[np.ndarray, np.ndarray]:
```
thomasZen: maybe call this split_data_in_train_dev_set

openretina/data_io/base_dataloader.py
```diff
@@ -369,3 +352,87 @@ def get_movie_dataloader(
...
+        """
+        Compute validation responses and updated train responses stripped from validation clips.
+
+        Returns:
+            Tuple[np.ndarray, np.ndarray]: The updated train and validation responses.
+        """
+        movie_ordering = np.arange(self.num_clips)
+
+        # Initialise validation responses
+        base_movie_sorting = np.argsort(movie_ordering)
```
thomasZen: I do not really understant what this does.
Aren't these two lines the same as:
base_movie_sorting = np.arange(self.num_clips)

openretina/data_io/base_dataloader.py
```diff
@@ -369,3 +352,87 @@ def get_movie_dataloader(
...
+            responses_val[i * self.clip_length : (i + 1) * self.clip_length, :] = self.responses_train_and_val[
+                grab_index * self.clip_length : (grab_index + 1) * self.clip_length,
+                :,
+            ]
+            validation_mask[
+                (grab_index * self.clip_length) : (grab_index + 1) * self.clip_length,
+                :,
+            ] = False
+
+        responses_train = self.responses_train_and_val[validation_mask].reshape(-1, self.num_neurons)
```
thomasZen: I think it's easier to implement this as:
```
val_idx_set = set(self.val_idx)
train_ar, val_ar = [], []
for i in range(self.num_clips):
    clp = self.responses_train_and_val[i * self.clip_length : (i + 1) * self.clip_length)
    if i in val_idx_set:
        val_ar.append(clp)
    else:
        train_ar_append(clp)
```

openretina/data_io/hoefling_2024/dataloaders.py
```diff
@@ -128,21 +129,17 @@ def extract_data_info_from_dataloaders(
...
 ):
-    assert all(field in next(iter(neuron_data_dictionary.values())) for field in ["responses_final", "stim_id"]), (
-        "Check the neuron data dictionary sub-dictionaries for the minimal"
-        " required fields: 'responses_final' and 'stim_id'."
-    )
     assert (
-        next(iter(neuron_data_dictionary.values()))["stim_id"] == 5
-    ), "This function only supports natural movie stimuli."
+        getattr(next(iter(neuron_data_dictionary.values())), "stim_id") == "natural"
+    ), f"This function only supports natural movie stimuli, not {next(iter(neuron_data_dictionary.values())).stim_id}."
```
thomasZen: In theory the stim_id might be different between these two, because we call next twice (also once we we use getattr, and once `.stim_id`.
maybe do sth like 
```
stim_ids = {x.stim_id for x in neuron_data_dictionary.values()}
if stim_ids != {"natural"}:
    raise ValueError(f"...., but where {stim_ids=})
```

openretina/data_io/hoefling_2024/responses.py
```diff
@@ -229,6 +189,50 @@ def map_to_range(self, max, val):
...
+
+        num_neurons = neural_responses.shape[0]
+
+        responses_test = np.zeros((5 * clip_length, num_neurons))
+        responses_train_and_val = np.zeros((num_clips * clip_length, num_neurons))
+
+        test_responses_by_trial = []
+
+        # Note: the hardcoded indices are the location of test clips in Hoefling 2024
+        for roi in range(num_neurons):
```
thomasZen: I'd rename it to i or idx, introducing roi here might be confusing

openretina/data_io/maheswaranathan_2023/dataloader.py
```diff
@@ -47,10 +48,10 @@ def multiple_movies_dataloaders(
...
     ), "The keys of neuron_data_dictionary and movies_dictionary should match."
 
-    assert all(field in next(iter(neuron_data_dictionary.values())) for field in ["responses_final", "stim_id"]), (
-        "Check the neuron data dictionary sub-dictionaries for the minimal"
-        " required fields: 'responses_final' and 'stim_id'."
-    )
+    # assert all(field in next(iter(neuron_data_dictionary.values())) for field in ["responses_final", "stim_id"]), (
+    #     "Check the neuron data dictionary sub-dictionaries for the minimal"
+    #     " required fields: 'responses_final' and 'stim_id'."
+    # )
```
thomasZen: remove if we do not use it anymore

thomasZen (commented):
I think this is an improvement, I'll take a deeper look tomorrow.

openretina/data_io/base.py
fededagos: Could be useful for checks, we could consider having a custom class for multiple sessions instead of the nested dictionary we have now.

openretina/data_io/base_dataloader.py
fededagos: Will refactor

openretina/data_io/base_dataloader.py
fededagos: Good spot, this is useless code. It is here because it has a function in the Hoefling version, but not here.

openretina/data_io/base_dataloader.py
fededagos: I will have a look to simplify this, you are totally right.

openretina/data_io/hoefling_2024/dataloaders.py
fededagos: I will fix it, thanks!

openretina/data_io/hoefling_2024/responses.py
fededagos: Will do (this comes from older code)

fededagos (commented):
Thanks for the comments, I will implement everything and then report again

configs/maheswaranathan_core_readout.yaml
```diff
@@ -27,7 +27,7 @@ trainer:
 
 dataloader:
   batch_size: null
-  num_workers: 4
+  num_workers: 3
```
thomasZen, line 30: directly set to 0 then?
```suggestion
  num_workers: 0
```

thomasZen (approved):
LGTM maybe directly set num_workers to 0 in all configs if you're sure that helps with speed (and reproduciblity)

configs/maheswaranathan_core_readout.yaml
fededagos: I will add it in a new PR with temporary notimplementederrors to warn the user if they try to increase it

PR #75, opened 2024-11-27
Dataloading and cycling fixes

fededagos:
- More flexibility in configs regarding dataloading
- LongCycler fixed at 0 workers, raising error otherwise.

### DIFF
--- a/openretina/data_io/cyclers.py
+++ b/openretina/data_io/cyclers.py
@@ -27,7 +27,7 @@ class LongCycler(torch.utils.data.IterableDataset):
     Cycles through a dictionary of data loaders until the loader with the largest size is exhausted.
     In practice, takes one batch from each loader in each iteration.
     Necessary for dataloaders of unequal size.
-    Note: iterable dataloaders as this one can lead to duplicate data when using multiprocessing
+    Note: iterable dataloaders as this one can lead to duplicate data when using multiprocessing.
     """
 
     def __init__(self, loaders: dict[str, DataLoader], shuffle: bool = True):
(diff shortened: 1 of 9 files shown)


openretina/data_io/cyclers.py
```diff
@@ -36,6 +36,10 @@ def __init__(self, loaders: dict[str, DataLoader], shuffle: bool = True):
         self.shuffle = shuffle
 
     def __iter__(self):
+        worker_info = torch.utils.data.get_worker_info()
+        if worker_info is not None:
+            raise NotImplementedError("LongCycler does not support multiple workers yet.")
```
thomasZen: We could allow num_workers == 1. But this also lead to slower training times, right?

tests/data_io/test_cyclers.py
```diff
@@ -58,10 +58,10 @@ def test_short_cycler(num_workers: int):
...
         (0, False),
         (0, True),
-        (1, False),
-        (1, True),
-        (4, False),
-        (4, True),
+        # (1, False),
+        # (1, True),
+        # (4, False),
+        # (4, True),
```
thomasZen, line 64: We could add a comment here that we currently only support num_workers==0

thomasZen (approved):
LGTM

openretina/data_io/cyclers.py
fededagos: I did not test this, but I cannot see how it could be faster yes

thomasZen: Theoretically, because the two processes (for training on gpu and for data loading) can run in parallel. But you're right, based on the benchmarks it seems the overhead is greater than the gain.

thomasZen: Hm, I changed my mind, let's allow one worker, too (I tried it and it works and speed is similar):
```suggestion
        worker_info = torch.utils.data.get_worker_info()
        if worker_info is not None and worker_info.num_workers > 1:
            raise NotImplementedError("LongCycler does not support multiple workers yet.")
```

PR #76, opened 2024-11-29
Add regularization parameters to core

thomasZen:
Adds group_sparsity and spatial laplace regularization functions.


fededagos (approved):
Looks good!

thomasZen:
@fededagos I set the spatial regularization weights to 0.0 for now, because that does not change behavior compared to before.
I think before setting them to a positive value again we should tune them ideally.

PR #77, opened 2024-12-04
Dev/paper training section refactoring

fededagos:
Work in progress refactoring for paper training section.

- Fixed GRUCoreReadout init
- Added LNP model


openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -0,0 +1,242 @@
...
+        stimuli_all_sessions["".join(session.split("/")[-1])] = MoviesTrainTestSplit(
+            train=train_video,
+            test=test_video,
+            stim_id=stim_type,
+        )
+
+    return stimuli_all_sessions
+
+
+def return_fix_movie_torch(screensize, im_ensemble, list_fixations):
```
thomasZen:
```suggestion
def return_fix_movie_torch(screensize: tuple[int, int], im_ensemble: np.ndarray, list_fixations: np.ndarray) -> torch.Tensor:
```

openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -0,0 +1,242 @@
...
+        x_or = x_or.astype(int)
+        y_or = y_or.astype(int)
+        blockstimulus[ymin : ymax + 1, xmin : xmax + 1, ifix] = im_ensemble[np.ix_(y_or, x_or)][
+            :, :, list_fixations[0, ifix] - 1
+        ]
+
+    return blockstimulus
+
+
+def get_ranges(tr_x, tr_y, Nxs, Nys, Nx, Ny, rx, ry):
```
thomasZen:
```suggestion
def get_ranges(tr_x: int, tr_y: int, Nxs: int, Nys: int, Nx: int, Ny: int, rx:np.ndarray, ry: np.ndarray) -> tuple[int, int, int, int, int, int]:
```

openretina/py.typed
thomasZen, line 1: This is because it then gets recognized as a typed package?
https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/#adding-pytyped

pyproject.toml
```diff
@@ -26,7 +26,7 @@ find = {exclude = ["tests", "scripts"]}
 dependencies = {file = ["requirements.txt"]}
 
 [tool.setuptools.package-data]
-"openretina" = ["data/stimuli/*.pickle"]
+"openretina" = ["data/stimuli/*.pickle", "py.typed"]
```
thomasZen: you can already remove the data/stimuli/*.pickle if you want

openretina/py.typed
fededagos: Exactly :)

tests/utils/test_nn_fabrik_model_loading.py
```diff
@@ -12,6 +13,10 @@
         ("2_ventral1_20201021", Center(target_mean=(0.0, 0.0))),
     ],
 )
+@pytest.mark.skipif(
+    condition=not check_server_responding("https://gin.g-node.org/eulerlab/rgc-natstim/"),
+    reason="GIN server unreachable.",
+)
```
thomasZen, line 19: this seems good. I also implemented retries with tenacy, so I hope we run into this issue less.

openretina/data_io/base_dataloader.py
```diff
@@ -243,13 +244,13 @@ def get_next_bound(value, bounds):
 
 
 def handle_missing_start_indices(
-    movies: dict | np.ndarray | torch.Tensor, chunk_size: int | None, scene_length: int | None, split: str
-):
+    movies: np.ndarray | torch.Tensor, chunk_size: int | None, scene_length: int | None, split: str
```
thomasZen:
```suggestion
    movie_length: int chunk_size: int | None, scene_length: int | None, split: str
```
I think this would make the function even easier to understand and to test.

openretina/data_io/base_dataloader.py
```diff
@@ -277,73 +278,85 @@ def get_chunking_interval(split_name):
...
-        if split == "test":
-            return {k: [0] for k in movies.keys()}
-        return {k: np.arange(0, movies[k].shape[1], interval).tolist() for k in movies.keys()}
-    else:
-        if split == "test":
-            return [0]
-        return np.arange(0, movies.shape[1], interval).tolist()
+    if split == "test":
+        return [0]
+    return np.arange(0, movies.shape[1], interval).tolist()
```
thomasZen: Make this function easier with the previous suggestion of having movie_length as an input:
```
if split == "test":
    return [0]
else:
    interval = get_chunking_interval(split)
    return np.arange(0, movies.shape[1], interval).tolist()
```
and remove the "test" case from the get_chunking_interval function

openretina/data_io/base_dataloader.py
```diff
@@ -427,3 +444,112 @@ def response_dict(self) -> dict:
...
+    # Initialise dataloaders
+    dataloaders: dict[str, Any] = {"train": {}, "validation": {}, "test": {}}
+
+    for session_key, session_data in tqdm(neuron_data_dictionary.items(), desc="Creating movie dataloaders"):
+        # Extract all data related to the movies first
+        num_clips = movies_dictionary[session_key].train.shape[1] // clip_length
+
+        # Draw validation clips based on the random seed
+        rnd = np.random.RandomState(seed)
+        val_clip_idx = list(rnd.choice(num_clips, num_val_clips, replace=False))
```
thomasZen: These should be read from the config and passed to this function, see:
https://github.com/open-retina/open-retina/blob/6033dcf0174c2fc525751c1ed16ccfaea20db3ea/configs/dataloader/hoefling_2024.yaml#L6

openretina/data_io/hoefling_2024/dataloaders.py
```diff
@@ -187,16 +187,20 @@ def natmov_dataloaders_v2(
...
+                movie=movie,
                 responses=neuron_data.response_dict[fold],
                 roi_ids=neuron_data.roi_ids,
                 roi_coords=neuron_data.roi_coords,
                 group_assignment=neuron_data.group_assignment,
-                scan_sequence_idx=neuron_data.scan_sequence_idx,
                 split=fold,
                 chunk_size=clip_chunk_sizes[fold],
-                start_indices=start_indices[fold],
+                start_indices=start_idxes,
```
thomasZen: use `start_idc`, using idxes as a short form for indices seems a bit off.

openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -0,0 +1,242 @@
...
+from openretina.data_io.base import MoviesTrainTestSplit
+from openretina.utils.h5_handling import load_dataset_from_h5
+
+
+def load_all_stimuli(
+    base_data_path: str | os.PathLike,
+    stim_type: Literal["fixationmovie", "frozencheckerflicker", "gratingflicker", "imagesequence"] = "fixationmovie",
+    normalize_stimuli: bool = True,
+    specie: Literal["mouse", "marmoset"] = "mouse",
+    downsampled_size: tuple = (60, 80),
```
thomasZen:
```suggestion
    downsampled_size: tuple[int, int] = (60, 80),
```

openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -0,0 +1,242 @@
...
+
+        # Check all files in the session path to identify relevant files
+        for recording_file in os.listdir(session_path):
+            full_path = os.path.join(session_path, recording_file)
+
+            if str(recording_file).endswith(
+                f"{stim_type}_data_extracted_stimuli_{downsampled_size[0]}_{downsampled_size[1]}.npz"
+            ):
+                npz_file = full_path
+                npz_file_found = True
```
thomasZen: get rid of `npz_file_found`. `npz_file_found = npz_file is not None` if I understand it correctly, so no need for it.

openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -0,0 +1,242 @@
...
+                train_videos.append(downsample(rearrange(train_snippet, "h w n -> 1 n h w")))
+            train_video = torch.cat(train_videos, dim=1).cpu().numpy()
+
+            np.savez_compressed(
+                f"{mat_file[:-4]}_extracted_stimuli_{downsampled_size[0]}_{downsampled_size[1]}.npz",
+                train_data=train_video,
+                test_data=test_video,
+            )
+        else:
+            continue  # Skip session if no relevant file found
```
thomasZen, line 93: Print something?

openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -0,0 +1,242 @@
...
+
+        if normalize_stimuli:
+            train_video = train_video - train_video.mean() / train_video.std()
+            test_video = test_video - train_video.mean() / train_video.std()
+
+        stimuli_all_sessions["".join(session.split("/")[-1])] = MoviesTrainTestSplit(
```
thomasZen:
```suggestion
        stimuli_all_sessions["".join(session.split("/")[-1])] = MoviesTrainTestSplit(
```
`"".join(session.split("/")[-1])` should be the same as `session.split("/")[-1]`.

openretina/utils/misc.py
```diff
@@ -210,3 +211,12 @@ def __exit__(self, exc_type, exc_value, traceback):
     def _noop(self, *args, **kwargs):
         """A no-op display function to suppress IPython outputs."""
         pass
+
+
+def check_server_responding(url):
```
thomasZen:
```suggestion
def check_server_responding(url) -> bool:
```

thomasZen (commented):
I'm strongly in favor of being explicit about the validation_indices, and not defining them based on a random seed.
For the natstim/hoefling dataset, they are defined in the config and should be passed to the relevant function. In case the other datasets need this, too, I'd be in favor to put them in the config, too (or to hardcode them somewhere else if they should always stay the same)

openretina/data_io/karamanlis_2024/stimuli.py
fededagos: good point

openretina/data_io/base_dataloader.py
fededagos: I agree in spirit, but with this function we can deal with training movies that are not assumed to contain the same clips, and as such if passing the validation clip indices these will have a different content in the different sessions. Also, you need to know in advance the number of clips to put it in the config, and again some numbers may not be possible if the training sequences have different length. I will include the option to pass the indices explicitly even here, but keep the option to draw them for flexibility. If you feel strongly against the random drawing option I can remove it entirely

thomasZen: okay, good point.

>  will include the option to pass the indices explicitly even here, but keep the option to draw them for flexibility.

Sounds good

openretina/data_io/base_dataloader.py
```diff
@@ -427,3 +444,120 @@ def response_dict(self) -> dict:
...
+    movies = {
+        "train": movie_train_subset,
+        "validation": movie_val,
+        "test": movie_test,
+        "val_clip_idx": val_clip_idx,
+    }
+
+    return movies
+
+
```
thomasZen, line 469: I think this should not be a function.
It's only used once and just putting it inline or letting the generate_movie_function directly return this dictionary (or an equivalent object) seems better.

openretina/data_io/base_dataloader.py
```diff
@@ -427,3 +444,120 @@ def response_dict(self) -> dict:
...
+    for session_key, session_data in tqdm(neuron_data_dictionary.items(), desc="Creating movie dataloaders"):
+        # Extract all data related to the movies first
+        num_clips = movies_dictionary[session_key].train.shape[1] // clip_length
+
+        if val_clip_indices is not None:
+            val_clip_idx = val_clip_indices
+        else:
+            # Draw validation clips based on the random seed
+            rnd = np.random.RandomState(seed)
+            val_clip_idx = list(rnd.choice(num_clips, num_val_clips, replace=False))
```
thomasZen, line 528: i'd argue for val_clip_idc instead of ...idx (but does not matter much)

openretina/data_io/karamanlis_2024/responses.py
```diff
@@ -0,0 +1,69 @@
...
+                mean_test_response = np.mean(testing_responses, axis=0)
+
+                training_responses = (
+                    rearrange(training_responses, "block time neurons -> neurons (block time)") / fr_normalisation
+                )
+
+            else:
+                continue  # Skip session if no relevant file found
+
+            responses_all_sessions["".join(session.split("/")[-1])] = ResponsesTrainTestSplit(
```
thomasZen: it still seems like "".join() will join individual character and you could get rid of this as `session.split("/")[-1]` is already a string

thomasZen: Ah, I think you changed it below but not here.

openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -0,0 +1,245 @@
...
+        x_or = torch.tensor(x_or, dtype=torch.int64).unsqueeze(0)  # type: ignore
+        y_or = torch.tensor(y_or, dtype=torch.int64).unsqueeze(1)  # type: ignore
+        blockstimulus[ymin : ymax + 1, xmin : xmax + 1, ifix] = im_ensemble[
+            y_or, x_or, list_fixations_tensor[0, ifix] - 1
+        ]
+
+    return blockstimulus
+
+
+def return_fix_movie(screensize, im_ensemble, list_fixations):
```
thomasZen:
```suggestion
def return_fix_movie(screensize: tuple[int, int], im_ensemble: np.ndarray, list_fixations: np.ndarray) -> np.ndarray:
```

openretina/models/linear_nonlinear.py
```diff
@@ -0,0 +1,97 @@
...
+    # Linear nonlinear Poisson
+    def __init__(
+        self,
+        in_shape: Int[tuple, "channel time height width"],
+        outdims: int,
+        smooth_weight=0.0,
+        sparse_weight=0.0,
+        smooth_regularizer="LaplaceL2norm",
+        laplace_padding=None,
+        nonlinearity="exp",
```
thomasZen:
```suggestion
        smooth_weight: float = 0.0,
        sparse_weight: float = 0.0,
        smooth_regularizer: str = "LaplaceL2norm",
        laplace_padding=None,
        nonlinearity: str = "exp",
```

openretina/models/linear_nonlinear.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            regularizer_config = dict(padding=laplace_padding)
+
+        regularizer_config = (
+            dict(padding=laplace_padding, kernel=self.kernel_size)
+            if smooth_regularizer == "GaussianLaplaceL2"
+            else dict(padding=laplace_padding)
+        )
+
+        self._smooth_reg_fn = regularizers.__dict__[smooth_regularizer](**regularizer_config)
```
thomasZen, line 52: This seems to be shady code :sweat_smile: If we can just initalize it with hydra and then pass the object it seems to be a better option.

openretina/models/linear_nonlinear.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def forward(self, x: Float[torch.Tensor, "batch channels t h w"], data_key=None, **kwargs):
+        out = self.inner_product_kernel(x)
+        out = self.nonlinearity(out)
+        out = rearrange(out, "batch neurons t 1 1 -> batch t neurons")
+        return out
+
+    def weights_l1(self, average=True):
```
thomasZen:
```suggestion
    def weights_l1(self, average: bool = True):
```

openretina/modules/core/base_core.py
```diff
@@ -136,25 +167,16 @@ def save_weight_visualizations(self, folder_path: str) -> None:
...
-    def __repr__(self) -> str:
-        s = f"{repr(super())} [{self.__class__.__name__} regularizers: "
-        ret = []
-        for attr in filter(lambda x: "gamma" in x or "skip" in x, dir(self)):
-            ret.append(f"{attr} = {getattr(self, attr)}")
-        return s + "|".join(ret) + "]\n"
+class DummyCore(Core):
+    """
+    A dummy core that does nothing. Used for readout only models, like the LNP model.
+    """
```
thomasZen, line 173: That's nice.

thomasZen (approved):
LGTM, I added a few smaller comments though if you find the time.

openretina/data_io/base_dataloader.py
fededagos: Created an issue to track this, it is not long to do but probably requires testing, so setting it as low priority for now but I will come back to it

openretina/models/linear_nonlinear.py
fededagos: Created #80 to track this

PR #78, opened 2024-12-05
Support downloading files directly from gin server

thomasZen:
The default config for training the core-readout models contains now urls to data on [GIN](https://gin.g-node.org/teulerlab/open-retina) (you can still define regular paths, though).
These urls are used to download the data, which gets stored in a cache directory defined in the config. The first time you run this, you'll have to wait a bit until the download finishs (you'll see a log message about this). Afterwards, the data from the cache directory is used.

To test this, run the following command to train the low res model on a small subset of the data:
```
./scripts/train_core_readout.py --config-path ~/GitRepos/open-retina/configs --config-name example_train_core_readout_low_res responses_path=https://gin.g-node.org/teulerlab/open-retina/raw/master/responses/eulerlab/rgc_natsim_subset_only_naturalspikes_2024-08-14.h5
```

### DIFF
--- a/configs/example_train_core_readout_low_res.yaml
+++ b/configs/example_train_core_readout_low_res.yaml
@@ -4,7 +4,7 @@ defaults:
   - dataloader@natmov_dataloader: hoefling_2024 # looks for default under "dataloader" and names it "natmov_dataloader"
 
 exp_name: example_core_readout_low_res_filtered
-cache_folder: "./cache_folder/"
+cache_folder: "./openretina_cache_folder/"
 movies_path: "https://gin.g-node.org/teulerlab/open-retina/raw/master/stimuli/eulerlab/rgc_natstim_18x16_joint_normalized_2024-01-11.pkl"
 responses_path: "https://gin.g-node.org/teulerlab/open-retina/raw/master/responses/eulerlab/rgc_natstim_2024-08-14.h5"
 save_folder: exp/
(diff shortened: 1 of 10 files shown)


configs/example_train_core_readout_low_res.yaml
```diff
@@ -4,9 +4,9 @@ defaults:
   - dataloader@natmov_dataloader: hoefling_2024 # looks for default under "dataloader" and names it "natmov_dataloader"
 
 exp_name: example_core_readout_low_res_filtered
-data_folder: "/gpfs01/euler/data/SharedFiles/projects/TP12/"
-movies_filename: "2024-01-11_movies_dict_8c18928.pkl"
-responses_filename: "2024-08-14_neuron_data_responses_484c12d_djimaging.h5"
+cache_folder: "./cache_folder/"
```
fededagos: I would still maybe keep this as "data_folder" or "data_store" or similar, and maybe have a different one for the models... I will try to find some default when unifying the training scripts and we can discuss again then

fededagos (approved):

configs/example_train_core_readout_low_res.yaml
thomasZen: Sounds good, this implementation is with hoefling data setup in mind. If the other data folders contain more files this might be a better solution.
I also renamed the default cache folder to openretina_cache_folder.

PR #83, opened 2024-12-12
Response gradient analysis

lhoefling:


openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+from copy import deepcopy
+import torch
+import numpy as np
+from openretina.insilico.stimulus_optimization.objective import SingleNeuronObjective
+
+class MeiAcrossContrasts(nn.Module):
+    """
+    PyTorch Module wrapper around a stimulus to allow tracking response gradients onto the stimulus.
+    """
+    def __init__(self, contrast_values: torch.TensorType, stim: torch.TensorType):
```
thomasZen:
```suggestion
    def __init__(self, contrast_values: torch.TensorType, stim: torch.TensorType):
```
If it's a torch tensor the type annotations is normally torch.Tensor (i don't know what TensorType actually is

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+                 ], dim=1),
+            self.stim.squeeze()
+        )
+
+
+def trainer_fn(mei_contrast_gen: MeiAcrossContrasts, model_neuron: SingleNeuronObjective, optimizer=optim.Adam, lr=1) -> tuple:
+    """
+    Trainer function for getting the gradient on the MEI with different contrasts
+    """
+    optimizer = optimizer(mei_contrast_gen.parameters(), lr=lr)
```
thomasZen: this is a bit dangerous, as both the thing you call and the value you assign it to have the same name (i.e. `optimizer = optimizer(...)`, I would call it `optimizer = optimizer_class(...)` or sth. similar

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+        return torch.mul(
+            torch.stack(
+                [torch.ones_like(self.stim[:, 0, ...]) * self.contrast_values[0],
+                 torch.ones_like(self.stim[:, 0, ...]) * self.contrast_values[1]
+                 ], dim=1),
+            self.stim.squeeze()
+        )
+
+
+def trainer_fn(mei_contrast_gen: MeiAcrossContrasts, model_neuron: SingleNeuronObjective, optimizer=optim.Adam, lr=1) -> tuple:
```
thomasZen:
```suggestion
def trainer_fn(mei_contrast_gen: MeiAcrossContrasts, model_neuron: SingleNeuronObjective, optimizer=optim.Adam, lr: float = 1.0) -> tuple[np.ndarray, np.ndarray]:
```

openretina/utils/plotting.py
```diff
@@ -245,6 +248,79 @@ def polar_plot_of_direction_of_motion_responses(
     plt.gca().set_xticklabels([f"{x}°" for x in sorted_directions])
 
 
+class ColorMapper:
+    def __init__(self, cmap_name, vmin=0, vmax=1):
```
thomasZen:
```suggestion
    def __init__(self, cmap_name: str, vmin: float = 0, vmax: float = 1):
```

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    Trainer function for getting the gradient on the MEI with different contrasts
+    """
+    optimizer = optimizer(mei_contrast_gen.parameters(), lr=lr)
+    loss = model_neuron.forward(mei_contrast_gen())
+    loss.backward()
+    grad_val = deepcopy(mei_contrast_gen.contrast_values.grad)
```
thomasZen: .grad can be None, at least the type checker thinks that.
To fix this, I think you can write:
```suggestion
    grad_val: torch.Tensor = deepcopy(mei_contrast_gen.contrast_values.grad)  # type: ignore
```

openretina/utils/plotting.py
```diff
@@ -245,6 +248,79 @@ def polar_plot_of_direction_of_motion_responses(
...
+        self.norm = plt.Normalize(vmin=self.vmin, vmax=self.vmax)
+
+    def get_color(self, number):
+        color = self.cmap(self.norm(number))
+        return color
+
+
+def plot_vector_field_resp_iso(x: np.ndarray, y: np.ndarray, gradient_dict: np.ndarray, 
+                               resp_dict: np.ndarray, normalize_response: bool = False,
+                               rc_dict: Dict[str, Any] = {},
```
thomasZen:
```suggestion
                               rc_dict: dict[str, Any] = {},
```
if you write dict instead of Dict, then you do not have to `from typing import Dict`. python change it at some point to have types like `dict`, `list`, `tuple`.

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    PyTorch Module wrapper around a stimulus to allow tracking response gradients onto the stimulus.
+    """
+    def __init__(self, contrast_values: torch.TensorType, stim: torch.TensorType):
+        """
+        Initialize the ResponseGradient class.
+
+        Parameters:
+        contrast_values (torch.TensorType): A torch Tensor with 2 scalar values that the MEI green and UV channels will be multiplied with.
+        stim (torch.TensorType): MEI stimulus as torch Tensor.
```
thomasZen:
```suggestion
        contrast_values (torch.Tensor): A torch Tensor with 2 scalar values that the MEI green and UV channels will be multiplied with.
        stim (torch.Tensor): MEI stimulus as torch Tensor.
```
Does it have to be an MEI stimulus or can it be any stimulus?

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+from openretina.insilico.stimulus_optimization.objective import SingleNeuronObjective
+
+class MeiAcrossContrasts(nn.Module):
+    """
+    PyTorch Module wrapper around a stimulus to allow tracking response gradients onto the stimulus.
+    """
+    def __init__(self, contrast_values: torch.TensorType, stim: torch.TensorType):
+        """
+        Initialize the ResponseGradient class.
+
```
thomasZen: I'm a bit against these comments as it's obvious from the name of the function, and this line does not add anything useful imo (feel free to overrule this opinion though :))

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+        Parameters:
+        contrast_values (torch.TensorType): A torch Tensor with 2 scalar values that the MEI green and UV channels will be multiplied with.
+        stim (torch.TensorType): MEI stimulus as torch Tensor.
+        """
+        super().__init__()
+        self.contrast_values = nn.Parameter(contrast_values, requires_grad=True)
+        self.stim = nn.Parameter(stim, requires_grad=False)
+
+    def forward(self):
```
thomasZen, line 26: Maybe add a comment here. Does this: multiply the stimulus for each channel with the contrast values?

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+        return torch.mul(
+            torch.stack(
+                [torch.ones_like(self.stim[:, 0, ...]) * self.contrast_values[0],
+                 torch.ones_like(self.stim[:, 0, ...]) * self.contrast_values[1]
```
thomasZen:
```suggestion
                 torch.ones_like(self.stim[:, 1, ...]) * self.contrast_values[1]
```
maybe there's a typo above

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+            norm_grid[i, j] = np.linalg.norm(grid[:, i, j])
+    return grid, resp_grid, norm_grid, green_contrast_values, uv_contrast_values
+
+
+def equalize_channels(mei: torch.Tensor,
+                      flip_green: bool = False) -> torch.Tensor:
+    '''
+    Scales the (green and UV) channels of an MEI such that they have equal norm,
+    and the scaled MEI has the same norm as the original MEI. Optionally flips sign of green channel
+    :param mei: torch.Tensor, shape (1, 2, 50, 18, 16)
```
thomasZen: If you call the argument stimulus instead of mei it's a bit more general (assuming you can pass any stimulus to this function)

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,110 @@
...
+    :return: equalized_mei: torch.Tensor, shape (1, 2, 50, 18, 16)
+    '''
+    green_chan = mei[0, 0]
+    uv_chan = mei[0, 1]
+    green_norm = torch.norm(green_chan)
+    uv_norm = torch.norm(uv_chan)
+    total_norm = torch.norm(mei)
+    green_factor = (total_norm/2)/green_norm
+    uv_factor = (total_norm/2)/uv_norm
+    equalized_mei = torch.zeros_like(mei).cuda()
```
thomasZen:
```suggestion
    equalized_mei = torch.zeros_like(mei, device=mei.device)
```
I think if you change the code like this it will work on both cpu and gpu (I think it's best if we strive to have no explicit .cuda() calls in our code except there is a good reason to have them :) )

thomasZen (commented):
Cool, this looks good.
~I did not approve yet because there might be a small bug in MeiAcrossContrasts.forward (see comment).~

Also, if you run `make fix-all` it will reformat the new code and afterwards the codestyle checks should pass.

I did not look at the notebook yet, though.

openretina/insilico/tuning_analyses/response_gradient.py
thomasZen: nvm, it does not matter and is not a bug.

thomasZen: Thinking about this, I think you can make this more general such that it works for len(contrast_values) > 2, too:
```
import torch
stim = torch.rand((2,3,5,7))
big_contrast_values = contrast_values.view([len(contrast_values)] + [1 for _ in range(len(stim.shape)-2)])
scaled_stim = big_contrast_values * stim
```
E.g. the code above works. It only makes sense if you want to support models with more or less channels in the response gradient analysis, though.

openretina/insilico/tuning_analyses/response_gradient.py
lhoefling: indeed very dangerous, thanks for catching this!

openretina/insilico/tuning_analyses/response_gradient.py
lhoefling: hm, I think it would be best to generalize this function in the next step to allow an arbitrary number of dimensions to be varied. I would make this a feature to add, and leave this basic function as is (i.e. tailored for 2 channels) for now

openretina/insilico/tuning_analyses/response_gradient.py
```diff
@@ -0,0 +1,119 @@
...
+    Trainer function for getting the gradient on the MEI with different contrasts
+    """
+    optimizer = optimizer_class(mei_contrast_gen.parameters(), lr=lr)
+    loss = model_neuron.forward(mei_contrast_gen())
+    loss.backward()
+    grad_val: torch.Tensor = deepcopy(mei_contrast_gen.contrast_values.grad)  # type: ignore
+    optimizer.zero_grad()
+    return grad_val.detach().cpu().numpy().squeeze(), loss.detach().cpu().numpy().squeeze()
+
+def get_gradient_grid(stim: torch.TensorType,
```
thomasZen:
```suggestion
def get_gradient_grid(stim: torch.Tensor,
```

openretina/utils/plotting.py
```diff
@@ -245,6 +248,79 @@ def polar_plot_of_direction_of_motion_responses(
...
+
+    def get_color(self, number):
+        color = self.cmap(self.norm(number))
+        return color
+
+
+def plot_vector_field_resp_iso(x: np.ndarray, y: np.ndarray, gradient_dict: np.ndarray,
+                               resp_dict: np.ndarray, normalize_response: bool = False,
+                               rc_dict: dict[str, Any] = {},
+                               cmap: str = "hsv") -> None:
```
thomasZen:
```suggestion
                               cmap: str = "hsv") -> plt.Figure:
```
This function actually returns a figure at the end

openretina/utils/plotting.py
```diff
@@ -245,6 +248,79 @@ def polar_plot_of_direction_of_motion_responses(
...
+
+    with mpl.rc_context(rc_dict):
+        fig = plt.figure()
+
+        # Create a contour plot with isoresponse lines
+
+        plt.contourf(X, Y, Z, levels=levels, cmap=cmap, zorder=200)  # Change cmap to the desired colormap
+        cont_lines = plt.contour(X,Y,Z, levels=levels, cmap='jet_r',zorder=300)
+        plt.gca().clabel(cont_lines, inline=True, fmt='%1.0f',
+                         levels = cont_lines.levels[::2], colors="k", fontsize=5, zorder=400)
```
thomasZen:
```suggestion
                         levels = list(cont_lines.levels)[::2], colors="k", fontsize=5, zorder=400)
```
Apparently, cont_lines.levels is an Iterable, so something you can iterate over in a for loop (like `range(10)`), but that you cannot necessarily index. That why the type checks complain with:
```
openretina/utils/plotting.py:304: error: Value of type "Iterable[float]" is not indexable
```

To avoid that type error, you can convert that iterable to a list.

thomasZen (approved):
Approved :)

I made suggestions to fix the 3 remaining lines the type checker complains about, and you'll have to run `make fix-all` once to format your files.

PR #84, opened 2024-12-12
Unified training script

fededagos:
Re-factored and adapted existing code to use a single unified training script on all three datasets we currently support.


openretina/modules/readout/factorised_gaussian.py
```diff
@@ -40,7 +40,9 @@ def __init__(
         """we train on the log var and transform to var in a separate step"""
         self.mask_mean = torch.nn.Parameter(data=torch.zeros(self.outdims, 2), requires_grad=True)
         self.mask_log_var = torch.nn.Parameter(data=torch.zeros(self.outdims), requires_grad=True)
-        self.grid = torch.nn.Parameter(data=self.make_mask_grid(outdims, w, h), requires_grad=False)
+
+        # Grid is fixed and untrainable, so we register it as a buffer
+        self.register_buffer("grid", self.make_mask_grid(outdims, w, h))
```
fededagos, line 45: This way the grid parameters also do not show up in the summary that lightning prints upon model creation.

openretina/data_io/hoefling_2024/stimuli.py
```diff
@@ -1,9 +1,16 @@
 import numpy as np
 import torch
 
+from openretina.data_io.base import MoviesTrainTestSplit
 from openretina.data_io.base_dataloader import generate_movie_splits
+from openretina.data_io.hoefling_2024.constants import CLIP_LENGTH, NUM_CLIPS
 
-from .constants import CLIP_LENGTH, NUM_CLIPS
+
+def movies_from_pickle(file_path: str):
```
fededagos, line 9: I needed to add this function for convenience, as with hydra you cannot call classmethods directly unfortunately

thomasZen:
I tested training and loading of a the low res model on cpu and it works.

Makefile
```diff
@@ -23,6 +23,7 @@ fix-formatting:
 fix-all: fix-codestyle fix-formatting
 
 test-corereadout:
-	./scripts/train_core_readout.py --config-path ../configs --config-name example_train_core_readout_low_res responses_path=https://gin.g-node.org/teulerlab/open-retina/raw/master/responses/eulerlab/rgc_natsim_subset_only_naturalspikes_2024-08-14.h5 +trainer.limit_train_batches=1 trainer.max_epochs=1 +trainer.limit_val_batches=1 +trainer.limit_test_batches=1 natmov_dataloader.batch_size=2
+	python ./scripts/train.py --config-path ../configs --config-name hoefling_2024_core_readout_low_res \
```
thomasZen, line 26:
```suggestion
	./scripts/train.py --config-path ../configs --config-name hoefling_2024_core_readout_low_res \
```
I think I changed it to the data repository

configs/hoefling_2024_core_readout_low_res.yaml
```diff
@@ -0,0 +1,43 @@
...
+  batch_size: 128
+
+
+model:
+  in_shape: [2, 150, 18, 16]
+
+data:
+  root_dir: "/Data/"
+  data_dir: ${data.root_dir}/euler_data/
+  output_dir: ${data.root_dir}/open-retina/runs
```
thomasZen: Hm, I'd prefer to put root_dir and output_dir in the root namespace. As these are not only data related but also contain the output of the model.
And can we change `/Data/` as most people will not have this directory on their local computer, maybe to sth like `openretina_assets`?

configs/karamanlis_2024_core_readout.yaml
```diff
@@ -0,0 +1,41 @@
+defaults:
+  - data_io: karamanlis_2024
+  - dataloader: karamanlis_2024
+  - model: GRU_core_readout
```
thomasZen: out of interest: did the GRU core readout work better for this model than the "default" core readout?

configs/data_io/hoefling_2024.yaml
```diff
@@ -0,0 +1,37 @@
+stimuli:
+  _target_: openretina.data_io.hoefling_2024.stimuli.movies_from_pickle
+  _convert_: object
+  file_path:
+    _target_: openretina.utils.file_utils.get_local_file_path
+    _convert_: object
+    file_path: ${data.movies_path}
```
thomasZen: Add a comment, sth. like: `# read from the main config as this can refer to high or low resolution videos`

configs/data_io/hoefling_2024.yaml
```diff
@@ -0,0 +1,37 @@
...
+    verbose: ${quality_checks.verbose}
+    all_responses:
+      # all_responses are loaded from the .h5 file
+      _target_: openretina.utils.h5_handling.load_h5_into_dict
+      _convert_: object
+      # path can be either remote or local so needs to be resolved
+      file_path:
+        _target_: openretina.utils.file_utils.get_local_file_path
+        _convert_: object
+        file_path: ${data.responses_path}
```
thomasZen, line 36: I'd mention the link here instead of in the main config as its always the same link.

openretina/data_io/karamanlis_2024/responses.py
```diff
@@ -16,12 +16,69 @@
...
+        stim_type (str): The stimulus type to filter files.
+        fr_normalisation (float): Normalization factor for firing rates.
+
+    Returns:
+        ResponsesTrainTestSplit | None: Loaded responses for the session or None if no relevant file found.
+    """
+    for recording_file in os.listdir(session_path):
+        full_path = os.path.join(session_path, recording_file)
+
+        if recording_file.endswith(f"{stim_type}_data.mat"):
```
thomasZen: Suggestion:
- First get all files: `recording_file_names = [x for x in os.listdir(session_path) if x.endswith(f"{stim_type}_data.mat")`
- Then check if it's exactly one relevant file (I'd actuatlly throw an exception instead of returning `None` as I'm a bit sceptical of silently not reading in sessions if we did not find the file):
```
if len(recording_file_names) == 0:
    raise IOError(f"No file with postfix {f'{stim_type}_data.mat'} found in folder {session_path}")  # or return None
elif len(recording_file_names) > 1:
    raise IOError(f"Multiple files with postifx {f'{stim_type}_data.mat'} found in folder {session_path}: {recording_file_name=}")

full_path = os.path.join(session_path, recording_file_names[0])
...
```

Then you get rid of two intendation levels (the for loop and the if recording_file.endswith) and directly do the validation at the beginning of the function.

openretina/data_io/karamanlis_2024/responses.py
```diff
@@ -36,34 +93,8 @@ def load_all_responses(
...
-
-            responses_all_sessions[session.split("/")[-1]] = ResponsesTrainTestSplit(
-                train=training_responses,
-                test=mean_test_response,
-                test_by_trial=testing_responses,
-                stim_id=stim_type,
-            )
+        responses = load_responses_for_session(session_path, stim_type, fr_normalization)
+        if responses:
+            responses_all_sessions[session.split("/")[-1]] = responses
```
thomasZen: adapt this if you go with exceptions

openretina/data_io/karamanlis_2024/stimuli.py
```diff
@@ -18,13 +18,103 @@
...
+
+        if recording_file.endswith(
+            f"{stim_type}_data_extracted_stimuli_{downsampled_size[0]}_{downsampled_size[1]}.npz"
+        ):
+            npz_file = full_path
+            break  # Prefer `.npz` file
+
+        elif recording_file.endswith(f"{stim_type}_data.mat"):
+            mat_file = full_path
+
```
thomasZen: Either add a comment already here why we prefer npz files over mat files, or refactor it towards more explicitly writing the code in this direction, e.g.:
```
npz_postfilx = f"{stim_type}_data_extracted_stimuli_{downsampled_size[0]}_{downsampled_size[1]}.npz"
`npz_file_names = `[x for x in os.listdir(session_path) if x.endswith(npz_postfix)]

if len(npz_file_names) == 1:
    # found already preproc file
    ...
else:
    # fall back to mat file and fail/return None if there's none
    ....
```
but ignore this if it doesn't feel right or if it's more work than I think.

openretina/data_io/maheswaranathan_2023/responses.py
```diff
@@ -5,7 +5,51 @@
...
+    The base data path should point to the location of neural_code_data/ganglion_cell_data
+    (See https://doi.org/10.25740/rk663dm5577 for dataset download)
+    """
+    responses_all_sessions = {}
+    for session in os.listdir(base_data_path):
+        session_path = os.path.join(base_data_path, session)
+        if not os.path.isdir(session_path):
+            continue
+        for recording_file in os.listdir(session_path):
+            if str(recording_file).endswith(f"{stim_type}.h5"):
```
thomasZen:
```suggestion
        for recording_file in [x for x in os.listdir(session_path) if x.endswith(f"{stim_type}.h5)]:
```
or if the cast is required str(x).endswith

openretina/data_io/maheswaranathan_2023/responses.py
```diff
@@ -5,7 +5,51 @@
...
+    Load all neural responses from sessions within subfolders in a given base data path.
+
+    The base data path should point to the location of neural_code_data/ganglion_cell_data
+    (See https://doi.org/10.25740/rk663dm5577 for dataset download)
+    """
+    responses_all_sessions = {}
+    for session in os.listdir(base_data_path):
+        session_path = os.path.join(base_data_path, session)
+        if not os.path.isdir(session_path):
+            continue
```
thomasZen:
```suggestion
    for session in [x.name for x in os.scandir(".") if x.is_dir()]:
        session_path = os.path.join(base_data_path, session)
```
this is shorter and scandir is more efficient according to the python docs: https://docs.python.org/3/library/os.html#os.scandir

openretina/data_io/maheswaranathan_2023/responses.py
```diff
@@ -5,7 +5,51 @@
...
+
+                # Load neural responses
+                train_session_data = load_dataset_from_h5(recording_file, f"/train/response/{response_type}")
+                test_session_data = load_dataset_from_h5(recording_file, f"/test/response/{response_type}")
+
+                assert (
+                    train_session_data.shape[0] == test_session_data.shape[0]
+                ), "Train and test responses should have the same number of neurons."
+
+                responses_all_sessions["".join(session.split("/")[-1])] = ResponsesTrainTestSplit(
```
thomasZen:
```suggestion
                responses_all_sessions[session] = ResponsesTrainTestSplit(
```
why does above not work. Session should be a filename, and not a path. Also avoiding the use of `/` and `\` for anything file related will make it easier to adapt to windows.

openretina/data_io/maheswaranathan_2023/stimuli.py
```diff
@@ -6,41 +6,35 @@
...
     The base data path should point to the location of neural_code_data/ganglion_cell_data
     (See https://doi.org/10.25740/rk663dm5577 for dataset download)
     """
-    responses_all_sessions = {}
     stimuli_all_sessions = {}
     for session in os.listdir(base_data_path):
-        # List sessions in the base data path
         session_path = os.path.join(base_data_path, session)
         if not os.path.isdir(session_path):
             continue
```
thomasZen: Use `for session in [x.name for x in os.scandir(".") if x.is_dir()]: ` to get rid of continue statement

openretina/data_io/maheswaranathan_2023/stimuli.py
```diff
@@ -51,30 +45,13 @@ def load_all_sessions(
                 test_video = test_video[None, ...]
 
                 if normalize_stimuli:
-                    train_video = train_video - train_video.mean() / train_video.std()
-                    test_video = test_video - test_video.mean() / test_video.std()
+                    train_video_preproc = (train_video - train_video.mean()) / train_video.std()
+                    test_video = (test_video - train_video.mean()) / train_video.std()
+                    train_video = train_video_preproc
```
thomasZen: maybe this should be a function (e.g. `def normalize_train_test(train, test) -> tuple[np.ndarray, np.ndarray]`), we do this in multiple places

openretina/models/core_readout.py
```diff
@@ -119,20 +122,35 @@ def save_weight_visualizations(self, folder_path: str) -> None:
         self.core.save_weight_visualizations(os.path.join(folder_path, "weights_core"))
         self.readout.save_weight_visualizations(os.path.join(folder_path, "weights_readout"))
 
+    def compute_readout_input_shape(
+        self, core_in_shape: tuple[int, int, int, int], core: Core | None = None
+    ) -> tuple[int, int, int, int]:
+        if core is None:
+            core = self.core
```
```suggestion
    def compute_readout_input_shape(
        self, core_in_shape: tuple[int, int, int, int], core: Core,
    ) -> tuple[int, int, int, int]:
```
either that or remove core from the arguments of the function and always use `self.core`

openretina/modules/core/base_core.py
```diff
@@ -86,14 +88,22 @@ def __init__(
         self._downsample_input_kernel_size = (
             list(downsample_input_kernel_size) if downsample_input_kernel_size is not None else None
         )
+        if self._cut_first_n_frames and not input_padding:
+            warnings.warn(
+                (
+                    "Cutting frames from the core output can lead to unexpected results if the input is not padded."
+                    "Double check the core output shape."
```
thomasZen:
```suggestion
                    f"{self._cut_first_n_frames=}, {input_padding=}. Double check the core output shape."
```
maybe that's useful for the user, too.

openretina/modules/core/base_core.py
```diff
@@ -86,14 +88,22 @@ def __init__(
...
+            )
 
         self._input_weights_regularizer_spatial = FlatLaplaceL23dnorm(padding=0)
 
         self.features = torch.nn.Sequential()
         for layer_id, (num_in_channels, num_out_channels) in enumerate(zip(channels[:-1], channels[1:], strict=True)):
             layer: dict[str, torch.nn.Module] = OrderedDict()
-            padding = "same"  # ((temporal_kernel_sizes[layer_id] - 1) // 2,
-            # (spatial_kernel_sizes[layer_id] - 1) // 2, (spatial_kernel_sizes[layer_id] - 1) // 2)
+            padding = "same" if input_padding and layer_id == 0 else ("same" if hidden_padding and layer_id != 0 else 0)
```
thomasZen, line 106: Padding can in theory be more nuanced for torch Convs:
```
padding controls the amount of padding applied to the input. It can be either a string {‘valid’, ‘same’} or an int / a tuple of ints giving the amount of implicit padding applied on both sides.
```

Why not let the user provide this value, so the type of input_padding and hidden_padding is `int | str` and we just write:
```
padding = input_padding if layer_id == 0 else hidden_padding
```
here (that requires changing the config though)

openretina/modules/layers/regularizers.py
```diff
@@ -137,7 +139,7 @@ def forward(self, x: torch.Tensor, avg: bool = False) -> torch.Tensor:
 
         oc, ic, k1, k2, k3 = x.size()
         assert (k2, k3) == (1, 1), "space dimensions must be one"
-        return agg_fn(self.laplace(x.view(oc * ic, 1, k1)).pow(2)) / agg_fn(x.view(oc * ic, 1, k1).pow(2))
+        return agg_fn(self.laplace(x.view(oc * ic, 1, k1)).pow(2)) / (agg_fn(x.view(oc * ic, 1, k1).pow(2)) + EPS)
```
thomasZen, line 142: valid point, did this drastically change something?

openretina/modules/readout/factorised_gaussian.py
thomasZen: Cool, that's great.

openretina/utils/file_utils.py
```diff
@@ -16,21 +18,37 @@ def optionally_download_from_url(
...
-        if response.status_code != 200:
-            raise FileNotFoundError(
-                f"Received status code {response.status_code} " f"when trying to download from {full_url=}"
-            )
-        with open(cache_path, "wb") as f:
-            f.write(response.content)
-        LOGGER.info(f"Stored contents of {full_url} in {cache_path}")
-    return cache_path
+    download_file_name = path.split("/")[-1]
+    target_download_path = Path(cache_folder) / Path(download_file_name)
```
thomasZen, line 22: I think this is dangerous, as it will download the urls paths `eulerlab/responses.h5` and `teslalab/responses.h5` to the same location.

I prefer the previous method.

openretina/utils/file_utils.py
```diff
@@ -16,21 +18,37 @@ def optionally_download_from_url(
...
+    os.makedirs(Path(target_download_path).parent, exist_ok=True)
+    if not os.path.exists(target_download_path):
+        full_url = urljoin(base_url, path)
+        LOGGER.info(f"Downloading {full_url} to {target_download_path}...")
+
+        with requests.get(full_url, stream=True) as response:
+            if response.status_code != 200:
+                raise FileNotFoundError(
+                    f"Received status code {response.status_code} " f"when trying to download from {full_url=}"
```
thomasZen:
```suggestion
                    f"Received status code {response.status_code} when trying to download from {full_url=}"
```

openretina/utils/file_utils.py
```diff
@@ -41,6 +59,8 @@ def get_local_file_path(file_path: str, cache_folder: str) -> Path:
             file_path[len(GIN_BASE_URL) :],
             os.path.expanduser(cache_folder),
         )
+    elif file_path.startswith("https://"):
+        raise NotImplementedError(f"Only {GIN_BASE_URL} data storage is supported for now.")
     elif file_path.startswith("http"):
         raise ValueError(f"Downloading from http {file_path=} is not supported yet.")
```
thomasZen: I think we should wrap these two cases in a single case (why separate http and https?)

scripts/train.py
```diff
@@ -0,0 +1,77 @@
...
+@hydra.main(version_base="1.3", config_path="../configs", config_name="hoefling_2024_core_readout_high_res")
+def main(cfg: DictConfig) -> None:
+    log.info("Logging full config:")
+    log.info(OmegaConf.to_yaml(cfg))
+
+    ### Import data
+    movies_dict = hydra.utils.call(cfg.data_io.stimuli)
+    neuron_data_dict = hydra.utils.call(cfg.data_io.responses)
+
+    if cfg.check_matches:
```
thomasZen:
```suggestion
    if cfg.check_matches_between_stimuli_and_responses:
```
Be more verbose (not sure if my proposal is perfect, though)

thomasZen (commented):
It's awesome to have a single train.py file.

I left many comments, but most of them are smallish (feel free to ignore the stylistic ones if you disagree or they are too much work)

Makefile
```diff
@@ -23,6 +23,7 @@ fix-formatting:
...
+	data.root_dir=./openretina_cache_folder data.responses_path=https://gin.g-node.org/teulerlab/open-retina/raw/master/responses/eulerlab/rgc_natsim_subset_only_naturalspikes_2024-08-14.h5 \
+	+trainer.limit_train_batches=1 trainer.max_epochs=1 +trainer.limit_val_batches=1 +trainer.limit_test_batches=1 dataloader.batch_size=2
```
thomasZen, line 29:
```suggestion
        exp_name: test_hoefling_2024_low_res \
	+trainer.limit_train_batches=1 trainer.max_epochs=1 +trainer.limit_val_batches=1 +trainer.limit_test_batches=1 dataloader.batch_size=2
```
add a name that makes it obvious that it's just a test. (I forgot this previously)

configs/data_io/hoefling_2024.yaml
fededagos: I kind of like the added flexibility of looking up into the main config. If the link changes (or if you want to specify a local path), you do not need to dig into these configs but can modify it only at the top level

configs/hoefling_2024_core_readout_low_res.yaml
fededagos: I think these are a bit arbitrary and should be user defined... Maybe a good way to enforce this could be:
```suggestion
  root_dir: ??? # Choose the location of the root directory for the project
  data_dir: ${data.root_dir}/Data/euler_data/ # Modify according to preferred data location
  output_dir: ${data.root_dir}/openretina_assets/runs 
```

configs/karamanlis_2024_core_readout.yaml
fededagos: I did not check... I added this to the todo runs before uploading checkpoints :)

openretina/data_io/karamanlis_2024/responses.py
fededagos: I will! I will however return a warning in the case where no file is found. in their data, certain sessions did not see all stimuli they had. I think it is more convenient to have the outer level function go over all sessions, and then maybe warn that certain sessions did not have the specified stimulus, compared to having to specify which sessions to load manually

openretina/data_io/karamanlis_2024/stimuli.py
fededagos: I will add a comment but not refactor it, as I think that compared to the code for the responses it would not speed up things substantially.

openretina/data_io/maheswaranathan_2023/responses.py
fededagos: Good catch, probably a copy/paster error from other functions.

openretina/data_io/maheswaranathan_2023/stimuli.py
fededagos: Good point, will add it in base.

openretina/models/core_readout.py
fededagos: I will make it required (this needs to be called before super.init(), so self.core will be undefined. )

openretina/modules/core/base_core.py
fededagos: I agree in principle. In practice, there are a few issues why I am hesitant to implement this currently:
- The GRU core has only true or false as input. This is suboptimal, and I think the GRUCore should inherit all base functionality from this base convolutional core. It is a bit of a more involved refactor though.
- It really requires "trusting" the user and ourselves to provide the right padding in the config. I think for this we should write an init function like "check_padding", which checks whether valid or same is provided, and if a tuple is provided parses it to see if it is compatible with the input tensor. Torch probably throws an error if you do not pass this correctly already, but I think then we should at least document well that we pass the padding directly "as is" to torch.
- WE could even have more nuanced hidden paddings, i.e. not the same for every hidden layer. In that case, writing up a "check_padding" or "parse_padding" function would become even more important.

openretina/modules/layers/regularizers.py
fededagos: No, I thought it was the reason while Linear-Nonlinear Poisson training was diverging but it was not... It will not hurt stability anyway :)

openretina/utils/file_utils.py
fededagos: I agree, but it will depend on the cache_folder the user provides, so (at least in my mind), it should be the user responsibility to provide the right target folder for the data (which right now in the configs is quite explicit).
I did not mind the nesting from GIN in principle, but it was a bit too much IMO...

configs/hoefling_2024_core_readout_low_res.yaml
thomasZen: Sounds reasonable. (that'd mean we'd have to call the example scripts with ...root_dir=xyz).

configs/model/GRU_core_readout.yaml
```diff
@@ -15,6 +15,18 @@ core_input_padding: false
...
+core_gru_kwargs:
+  rec_channels: 64
+  input_kern: 4
+  rec_kern: 4
+  groups: 1
+  gamma_rec: 1
+  pad_input: true
+
+# readout
+n_neurons_dict: ???
```
thomasZen: add a comment that this is set dynamically in the code

configs/model/base_core_readout.yaml
```diff
@@ -0,0 +1,31 @@
...
+core_gamma_in_sparse: 0.0
+core_gamma_hidden: 0.0
+core_gamma_temporal: 40.0
+core_input_padding: false
+core_hidden_padding: true
+maxpool_every_n_layers: null
+downsample_input_kernel_size: null
+
+# readout
+n_neurons_dict: ??? # needs to be set before training
```
thomasZen: maybe change comment to this is set dynamically in the code (if that's true)

openretina/data_io/karamanlis_2024/responses.py
thomasZen: maybe double check here if you need the `split("/")` call

openretina/modules/core/base_core.py
thomasZen: ok, let's keep it as is.

openretina/utils/file_utils.py
thomasZen: yeah, it was a bit much. But I think it can become annoying for new users to debug this and change the data folder themselves. Maybe the easy fix is to keep that in mind and try to make sure that at least the files we upload to gin have unique names.

thomasZen (approved):
Looks good, I left a few smaller comments though.

It might make sense to check if the different training runs still work before merging this (then it's easier to quickly fix stuff without requesting another pr review)

configs/hoefling_2024_core_readout_low_res.yaml
fededagos: Exactly, either data or having local copies :)

PR #85, opened 2024-12-22
Re-add readout initialisation

fededagos:
With the last PR, I removed (by mistake) the readout feature weights initialisation, which was setting the weights to a random normal with mean centered around 1/n_channels.

While I did this by mistake, I was surprised to see that this was enough to make training consistently diverge within the first few gradient descent steps with both the GRU and classic core architecture (and this is how I actually found out I deleted this by mistake).


thomasZen (approved):
Good catch.
Maybe call the PR title "Re-add readout initialisation"

PR #87, opened 2025-01-10
Adapt visualization script to up to three color channels

thomasZen:
Works for all pretrained models now. Example command:
```
./scripts/visualize_model_neurons.py --model_path models/hoefling_2024_GRU_high_res_model.ckpt --save_folder vis/hoefling_2024_GRU_high_res_model
```

Changes:
- Make plotting and video generation work for grey-scale videos
- Optionally infer stimulus shape from filename [1]
- Update visualize_model_neurons script to work with all models

[1] It might be useful to store the input shape in the CoreReadoutModel, too (we currently pass it to the constructor, but do not store it their). If doing so, we would not have to pass it to the visualize_model_neurons script.

### DIFF
--- a/scripts/visualize_model_neurons.py
+++ b/scripts/visualize_model_neurons.py
@@ -54,7 +54,7 @@ def parse_args():
         "--stimulus_shape",
         nargs="+",
         type=int,
-        default=[2, 50, 72, 64],
+        default=None,
         help="Stimulus shape: [color_channels, time_dim, height, width",
     )
 
@@ -109,12 +109,8 @@ def main(
     device: str,
     model_id: int,
     is_hoefling_ensemble_model: bool,
-    stimulus_shape: tuple[int, ...],
+    stimulus_shape: tuple[int, ...] | None,
 ) -> None:
-    if len(stimulus_shape) != 4:
-        raise ValueError(f"Invalid stimulus shape, needs to contain 4 integers, but was {stimulus_shape=}")
-    stimulus_shape = (1,) + tuple(stimulus_shape)
-
     model = load_model(
         model_path,
         device=device,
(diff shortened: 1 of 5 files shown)


scripts/visualize_model_neurons.py
```diff
@@ -21,34 +22,39 @@
...
     )
-    parser.add_argument("--model_path", default=DEFAULT_ENSEMBLE_MODEL_PATH)
     parser.add_argument(
         "--model_id",
         type=int,
         default=-1,
         help="If >= 0 load the ensemble model with that model_id, else use torch.load to load the model",
     )
-    parser.add_argument("--core_readout_lightning", action="store_true")
+    parser.add_argument("--is_hoefling_ensemble_model", action="store_true")
```
thomasZen, line 52: I changed this because I think loading the model into the core_readout model should be the default now.

fededagos:
> Works for all pretrained models now. Example command:
> 
> ```
> ./scripts/visualize_model_neurons.py --model_path models/hoefling_2024_GRU_high_res_model.ckpt --save_folder vis/hoefling_2024_GRU_high_res_model
> ```
> 
> Changes:
> 
>     * Make plotting and video generation work for grey-scale videos
> 
>     * Optionally infer stimulus shape from filename [1]
> 
>     * Update visualize_model_neurons script to work with all models
> 
> 
> [1] It might be useful to store the input shape in the CoreReadoutModel, too (we currently pass it to the constructor, but do not store it their). If doing so, we would not have to pass it to the visualize_model_neurons script.

Regarding [1]: I was thinking about this already regarding the video normalisation, and now this is one more strong reason to include input information directly in the model metadata. On top of this, do you think we should include also the values of the video used in normalisation? I noticed they are used in a couple of functions in the script.
And should we make these changes now or incorporate this fix for the time being and deal with these problems later?

fededagos (approved):
Looks good for now, I am merging. I have created issue #89 to track the video size and normalisation data, which we should include in the model. I would set this to high priority as soon as we are done writing the paper.

thomasZen:
> Looks good for now, I am merging. I have created issue https://github.com/open-retina/open-retina/issues/89 to track the video size and normalisation data, which we should include in the model. I would set this to high priority as soon as we are done writing the paper.

Sounds good. Thanks for writing up the issue.

PR #90, opened 2025-01-16
Most discriminative stimuli notebook

thomasZen:
- Option to specify in which file format to save visualizations
- Move CaptureOutputAndWarnings into it's own file (importing `from IPython.core.interactiveshell import InteractiveShell` was incredibly slow on the CIN cluster)
- Initialize the stimuli with variance 0.1 instead of 1 in visualization script
- Rename SingleNeuronObjective to ObjectiveIncrease and make it work for both single and multiple neurons
- Added a MDS notebook
- Updated the gradient analysis notebook as I renamed the SingleNeuronObjective and that notebook used the objective


fededagos (approved):
Everything looks good to me!

PR #91, opened 2025-01-21
First initial CLI attempt

fededagos:
Creating a CLI for the package scripts for easier use. 

@thomasZen What is your opinion on this?

As we include more scripts, we can just add them to the CLI so they are easily callable, even if someone installs the repo via pip and does not clone the git repository.

Changes by Thomas:
- Include visualize in CLI
- Throw away other scripts (I think we don't need them anymore)
- Try to do cli module structure more like https://packaging.python.org/en/latest/guides/creating-command-line-tools/
- When downloading file, create one subdirectory in openretina_cache (e.g. downloading lara's ensemble model downloads 20 files, the openretina_cache directory would become too crowded without any subdirectory)
- Put openretina_cache by default in `~/openretina_cache`. This is similar to what stanza is doing: https://stanfordnlp.github.io/stanza/download_models.html (you can overwrite this by `export OPENRETINA_CACHE = "your_path"`


thomasZen:
I like the idea, we could also add the visualization script as a command line tool (but I'm not sure if it should be it's own or a dedicated one).
If it's a dedicated one it should be straight forward to add it, e.g. with setup tools by adding it to the field scripts: https://docs.python.org/3.11/distutils/setupscript.html#installing-scripts (I did not check how to do it in our current setup yet, but it should be similar)

fededagos:
@thomasZen yes! The idea is to add all scripts that we can run easily on different model configurations, but have them all callable with `openretina ...` so like `openretina train ...` and `openretina visualize_activity ...` and so on. This would be handled under openretina/cli which simply redirects the calls to the appropriate script. I think this is a bit more elegant and perhaps more intuitive for users. Also with this you can make it so that by calling `openretina` you see all the available scripts you can run quite easily

thomasZen (approved):
LGTM

PR #92, opened 2025-01-22
Adding data_info field to models

### LINKED ISSUE #89: Include video size and normalisation information in model metadata
> Works for all pretrained models now. Example command:
> 
> ```
> ./scripts/visualize_model_neurons.py --model_path models/hoefling_2024_GRU_high_res_model.ckpt --save_folder vis/hoefling_2024_GRU_high_res_model
> ```
> 
> Changes:
> 
>     * Make plotting and video generation work for grey-scale videos
> 
>     * Optionally infer stimulus shape from filename [1]
> 
>     * Update visualize_model_neurons script to work with all models
> 
> 
> [1] It might be useful to store the input shape in the CoreReadoutModel, too (we currently pass it to the constructor, but do not store it their). If doing so, we would not have to pass it to the visualize_model_neurons script.

Regarding [1]: I was thinking about this already regarding the video normalisation, and now this is one more strong reason to include input information directly in the model metadata. On top of this, do you think we should include also the values of the video used in normalisation? I noticed they are used in a couple of functions in the script.
And should we make these changes now or incorporate this fix for the time being and deal with these problems later?

_Originally posted by @fededagos in https://github.com/open-retina/open-retina/issues/87#issuecomment-2587350963_

---

The following feature is necessary for several reasons:
1. Easier access to video size metadata for MEI generation
2. Consistent normalisation of models trained on more than one video → video normalisation mean and std need to be made hyperparameters and stored inside the model
3. Easier (and necessary) access to video normalisation values when evaluating the model on an unseen clip (of which the normalisation should follow what was used in training)

fededagos:
Resolves #89 by adding the "data_info" field to models and a function to compute data_info from the same information that is fed into the dataloader-generating function.

### DIFF
--- a/openretina/data_io/base.py
+++ b/openretina/data_io/base.py
@@ -53,9 +53,9 @@ class ResponsesTrainTestSplit:
     session_kwargs: dict[str, Any] = field(default_factory=lambda: {})
 
     def __post_init__(self):
-        assert (
-            self.train.shape[0] == self.test.shape[0]
-        ), "Train and test responses should have the same number of neurons."
+        assert self.train.shape[0] == self.test.shape[0], (
+            "Train and test responses should have the same number of neurons."
+        )
         if self.train.shape[0] > self.train.shape[1]:
             warnings.warn(
                 "The number of neurons is greater than the number of timebins in the train responses. "
--- a/openretina/data_io/base.py
+++ b/openretina/data_io/base.py
@@ -96,6 +96,15 @@ def normalize_train_test_movies(
     """
     z-score normalization of train and test movies using the mean and standard deviation of the train movie.
 
+    Parameters:
+    - train: train movie with shape (channels, time, height, width)
+    - test: test movie with shape (channels, time, height, width)
+
+    Returns:
+    - train_video_preproc: normalized train movie
+    - test_video_preproc: normalized test movie
+    - norm_stats: dictionary containing the mean and standard deviation of the train movie
+
     Note: The functions casts the input to torch tensors to calculate the mean and standard deviation of large
     inputs more efficiently.
     """
--- a/openretina/data_io/base.py
+++ b/openretina/data_io/base.py
@@ -53,9 +53,9 @@ class ResponsesTrainTestSplit:
     session_kwargs: dict[str, Any] = field(default_factory=lambda: {})
 
     def __post_init__(self):
-        assert self.train.shape[0] == self.test.shape[0], (
-            "Train and test responses should have the same number of neurons."
-        )
+        assert (
+            self.train.shape[0] == self.test.shape[0]
+        ), "Train and test responses should have the same number of neurons."
         if self.train.shape[0] > self.train.shape[1]:
             warnings.warn(
                 "The number of neurons is greater than the number of timebins in the train responses. "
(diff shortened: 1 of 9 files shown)


openretina/data_io/base.py
```diff
@@ -84,7 +88,11 @@ def get_n_neurons_per_session(responses_dict: dict[str, ResponsesTrainTestSplit]
...
     train: Float[np.ndarray, "channels train_time height width"],
     test: Float[np.ndarray, "channels test_time height width"],
-) -> tuple[Float[np.ndarray, "channels train_time height width"], Float[np.ndarray, "channels test_time height width"]]:
+) -> tuple[
+    Float[np.ndarray, "channels train_time height width"],
+    Float[np.ndarray, "channels test_time height width"],
+    dict[str, float | None],
+]:
     """
     z-score normalization of train and test movies using the mean and standard deviation of the train movie.
```
thomasZen, line 98: Returns the normalized data for the train and the test data, and a dictionary with the mean and std of the train movie?
^ Add sth like this.

openretina/data_io/base.py
```diff
@@ -84,7 +88,11 @@ def get_n_neurons_per_session(responses_dict: dict[str, ResponsesTrainTestSplit]
 def normalize_train_test_movies(
     train: Float[np.ndarray, "channels train_time height width"],
     test: Float[np.ndarray, "channels test_time height width"],
-) -> tuple[Float[np.ndarray, "channels train_time height width"], Float[np.ndarray, "channels test_time height width"]]:
+) -> tuple[
+    Float[np.ndarray, "channels train_time height width"],
+    Float[np.ndarray, "channels test_time height width"],
+    dict[str, float | None],
```
thomasZen, line 94:
```suggestion
    dict[str, float],
```

Don't we always return a float here?

openretina/modules/readout/multi_readout.py
```diff
@@ -74,6 +74,10 @@ def save_weight_visualizations(self, folder_path: str) -> None:
             os.makedirs(readout_folder, exist_ok=True)
             self._modules[key].save_weight_visualizations(readout_folder)
 
+    @property
+    def sessions(self) -> list[str]:
+        return self.readout_keys()
+
```
thomasZen, line 80: We are not good at naming this. Would be best to have this consistent (either always sessions, or always readout_keys).

openretina/models/core_readout.py
```diff
@@ -181,6 +183,7 @@ def __init__(
         dropout_rate: float = 0.0,
         maxpool_every_n_layers: Optional[int] = None,
         downsample_input_kernel_size: Optional[tuple[int, int, int]] = None,
+        data_info: dict[str, int] | None = None,
```
```suggestion
        data_info: dict[str, Any] | None = None,
```

I think the int is wrong in the type annotation

openretina/models/core_readout.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         learning_rate: float,
         loss: nn.Module | None = None,
         correlation_loss: nn.Module | None = None,
+        data_info: dict[str, int] | None = None,
```
```suggestion
        data_info: dict[str, Any] | None = None,
```

I think the int is wrong in the type annotation

openretina/data_io/base.py
thomasZen: but I see now why you did it. If it's much work to get the type annotations consistent in lateron code you can keep it like that

thomasZen (approved):
Looks reasonable, and will make things easier :)

Joining the documentation with type enforcements for the data info object might be good. Maybe something like using https://docs.python.org/3/library/typing.html#typing.TypedDict gives us both, and it says at runtime it's also just a dict.

Anyways, I'll approve and you can decide if that's worth to change.

PR #93, opened 2025-01-23
Proposal for minimal example for README

thomasZen:
Changes:
- Import important classes already in `__init__` files
- Function to load main core readout models from remote
- Simplify some plotting functions
- Separate unit and end-to-end tests (for speed, as end to end tests take the longest time as they download a lot of data)
- Simple readme example:

```python
import torch
from openretina.models import *

model = load_core_readout_from_remote("hoefling_2024_base_low_res", "cpu")
responses = model.forward(torch.rand(model.stimulus_shape(time_steps=50)))
```

It's similarly easy to visualize weights now:
```python
from openretina.models import *
model = load_core_readout_from_remote("hoefling_2024_base_low_res", "cpu")

# Visualize a specific channel weight for the first convolutional layer
convolution_layer_fig = model.core.plot_weight_visualization(layer=0, in_channel=1, out_channel=0)

# Visualize the readout weights of a neuron for a particular readout session
session_key = model.readout.readout_keys()[0]
readout_fig = model.readout[session_key].plot_weight_for_neuron(5)

### CODE REVIEW

File: openretina/models/linear_nonlinear.py [hunk 1]
```diff
@@ -41,11 +41,6 @@ def __init__(
...
         nn.init.xavier_normal_(self.inner_product_kernel.weight.data)
 
-        if smooth_regularizer == "GaussianLaplaceL2":
-            regularizer_config = dict(padding=laplace_padding, kernel=self.kernel_size)
-        else:
-            regularizer_config = dict(padding=laplace_padding)
-
         regularizer_config = (
             dict(padding=laplace_padding, kernel=self.kernel_size)
             if smooth_regularizer == "GaussianLaplaceL2"
```
thomasZen, line 46: The deleted code did the same as the three lines below it, and thus did not have an effect.

README.md
```diff
@@ -17,8 +17,16 @@ For normal usage:
 pip install openretina
 ```
 
-For development:
+Test openretina by downloading a model and running a forward path
```
fededagos:
```suggestion
Test openretina by downloading a model and running a forward pass
```

README.md
```diff
@@ -17,8 +17,16 @@ For normal usage:
...
+```python
+import torch
+from openretina.models import *
 
+model = load_core_readout_from_remote("hoefling_2024_base_low_res", "cpu")
+responses = model.forward(torch.rand(model.stimulus_shape(time_steps=50)))
+```
+
+For development and to have download Jupyter notebooks:
```
fededagos:
```suggestion
For development and to have access to Jupyter notebooks:
```

openretina/models/core_readout.py
```diff
@@ -135,24 +145,9 @@ def compute_readout_input_shape(
...
-                return 2, default_time_dim, 72, 64
-            else:
-                return 2, default_time_dim, 18, 16
-        elif "maheswaranathan" in file_name:
-            return 1, default_time_dim, 50, 50
-        elif "karamanlis" in file_name:
-            return 1, default_time_dim, 60, 80
-        else:
-            raise ValueError(f"Could not infer stimulus shape from {file_name=}")
+    def stimulus_shape(self, time_steps: int, num_batches: int = 1) -> tuple[int, int, int, int, int]:
```
fededagos, line 148: Amazing util :)

openretina/models/linear_nonlinear.py
fededagos: Good catch

openretina/modules/readout/factorised_gaussian.py
```diff
@@ -124,26 +126,41 @@ def __repr__(self) -> str:
...
 
-    def save_weight_visualizations(self, folder_path: str) -> None:
-        masks = self.get_mask().detach().cpu().numpy()
-        mask_abs_max = np.abs(masks).max()
-        features = self.features.detach().cpu().numpy()
-        features_min = float(features.min())
-        features_max = float(features.max())
-        for neuron_id in range(masks.shape[0]):
-            mask_neuron = masks[neuron_id, :, :]
+    @functools.lru_cache(1)
```
fededagos: Did you test lru_cache in this context? I had vague memories it was not great when applied to classes and instance methods, in that it can lead unexpected results and potentially memory leaks. e.g. see this: http://rednafi.com/python/lru_cache_on_methods/
It could be fine though in our context I guess? Just potentially worth testing a bit

openretina/utils/plotting.py
```diff
@@ -39,12 +39,14 @@ def undo_video_normalization(
 
 
 def save_stimulus_to_mp4_video(
```
fededagos, line 41: Just as a comment for myself mostly, in the new PR I am working on for the ROI visualization, I added moviepy as a dependency because it sped up substantially the rendering and displaying of videos. I will probably need to check if it also saves time on this function (as compared to the cv2 rendering).

openretina/modules/readout/factorised_gaussian.py
thomasZen: I tested it so far with the small examples and ran the visualize_neurons script, and it seems to work.
It's a good point though and I will keep this in mind.
I added it here to not need one `.numpy()` call per neuron when plotting all neurons of a readout layer.

fededagos (approved):
Mostly tiny comments. I would check on lru_cache before merging though. If the speed up is not crazy, I would remove it just to stay on the safer side

thomasZen:
> I would check on lru_cache before merging though. If the speed up is not crazy, I would remove it just to stay on the safer side

I checked: at least on cpu there is no speedup (maybe it's different on gpu as there's actually a memory copy involved, but probably it's not the bottleneck). I removed it again.

PR #95, opened 2025-01-31
Dev: Fix types after pytorch 2.6 was released

thomasZen:
Pytorch 2.6 was released 2 days ago: https://pytorch.org/blog/pytorch2-6/

This apparently changed the types a bit, the patterns I encountered are that buffers are typed differently (I think it was `torch.Tensor | torch.nn.Module`) and that torch.nn.Module now is considered in types.

I fixed most type annotations inconsistencies with adding `# type: ignore` for now.
I fixed a genuine error in the description of a ValueError.


fededagos (approved):
All good :)

PR #96, opened 2025-01-31
Include visualize in cli

thomasZen:
- Include visualize in CLI
- Throw away other scripts (I think we don't need them anymore)
- Try to do cli module structure more like https://packaging.python.org/en/latest/guides/creating-command-line-tools/
- When downloading file, create one subdirectory in openretina_cache (e.g. downloading lara's ensemble model downloads 20 files, the openretina_cache directory would become too crowded without any subdirectory)
- Put openretina_cache by default in `~/openretina_cache`. This is similar to what stanza is doing: https://stanfordnlp.github.io/stanza/download_models.html (you can overwrite this by `export OPENRETINA_CACHE = "your_path"`


openretina/cli/__init__.py
```diff
@@ -1,55 +0,0 @@
-import argparse
```
thomasZen, line 1: moved to openretina/cli/openretina.py

fededagos (approved):
Looks good, I agree with all the changes. Will make things more organised and easy to extend. It is also nice to be able to set an environment variable for the cache folder

PR #97, opened 2025-02-01
Cmd options with dashes instead of underscore, set data.root_dir to OPENRETINA_CACHE_DIR if not provided

thomasZen:
- Dashes instead of underscores is recommended by [GNU](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html):
```
Option names are typically one to three words long, with hyphens to separate words.
```
- I think data.root_dir=OPENRETINA_CACHE_DIR is a reasonable option, as pretrained models are also downloaded to that directory. We can also include an easier example in the quickstart section then.
- Cache `~/openretina_cache` in the end-to-end tests and in unit-tests. This results in a speed up of the tests, the end to end tests now only need 2 minutes to run.

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -20,6 +20,12 @@ jobs:
           python-version: "3.10"
           cache: 'pip'
 
+      - name: Load Openretina_cache
+        uses: actions/cache@v4
+        with:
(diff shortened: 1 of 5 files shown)


fededagos (approved):
LGTM, the cache is a great addiiton to the unit tests :)

PR #98, opened 2025-02-05
Training section notebooks

fededagos:


openretina/cli/train.py
```diff
@@ -23,6 +23,9 @@ def train_model(cfg: DictConfig) -> None:
     log.info("Logging full config:")
     log.info(OmegaConf.to_yaml(cfg))
 
+    ### Set cache folder
+    os.environ["OPENRETINA_CACHE_FOLDER"] = cfg.data.root_dir
```
thomasZen, line 27: I'm a bit skeptical of the current layout as in the configs we use `data.root_dir` both for storing data and training logs and checkpoints. I think we should make a bigger difference between these two locations.

openretina/eval/metrics.py
```diff
@@ -192,7 +192,7 @@ def explainable_vs_total_var(
     return var_ratio, explainable_var
 
 
-def fev(
+def feve(
```
thomasZen, line 195: maybe just call that fraction_of_explainable_variance?

README.md
```diff
@@ -1,4 +1,4 @@
-# OpenRetina
+# OpenRetina <img src="assets/openretina_logo.png" align="right" width="120" />
```
thomasZen, line 1: Nice :)

openretina/utils/file_utils.py
```diff
@@ -26,48 +26,56 @@ def optionally_download_from_url(
 ) -> Path:
     if cache_folder is None:
         cache_folder = get_cache_directory()
+
+    cache_folder = Path(cache_folder)
     download_file_name = Path(*Path(path).parts[-max_num_directories_in_cache_folder:])
+
+    # Search for the file in any subdirectory of cache_folder
+    existing_files = list(cache_folder.rglob(download_file_name.name))
+    if existing_files:
```
thomasZen:
```suggestion
    if len(existing_files) > 0:
```
Nitpicky, but I prefer this as a general pattern to be more specific.

openretina/utils/file_utils.py
```diff
@@ -26,48 +26,56 @@ def optionally_download_from_url(
...
+                for chunk in response.iter_content(chunk_size=chunk_size):
+                    f.write(chunk)
+                    progress_bar.update(len(chunk) // chunk_size)
+
+        LOGGER.info(f"Completed download of {full_url} to {target_download_path.resolve()}.")
+
+    except KeyboardInterrupt:
+        if target_download_path.exists():
+            target_download_path.unlink()
+            LOGGER.info(f"Partially downloaded file removed: {target_download_path}")
```
thomasZen:
```suggestion
            LOGGER.info(f"Partially downloaded file removed due to KeyboardInterrupt: {target_download_path}")
```

openretina/utils/file_utils.py
```diff
@@ -26,48 +26,56 @@ def optionally_download_from_url(
...
+        raise
+
+    except Exception as e:
+        LOGGER.error(f"Download failed for {full_url}. Error: {e}")
```
thomasZen:
```suggestion
        LOGGER.exception(f"Download failed for {full_url}:")
```
LOGGER.exception prints the whole stacktrace of the current exception, which is more informative (f"{e}" will only print the sth. like "ValueError" without the stacktrace I think)

openretina/data_io/hoefling_2024/constants.py
```diff
@@ -64,3 +64,4 @@
 CLIP_LENGTH = 150  # Frames
 
 STIMULI_IDS = {"natural": 5, "chirp": 1, "mb": 2}
+BADEN_TYPE_BOUNDARIES = [9, 14, 20, 28, 32]
```
thomasZen, line 67: Are these still used in notebooks or do we always use `RGC_GROUP_GROUP_ID_TO_CLASS_NAME`?

openretina/data_io/hoefling_2024/responses.py
```diff
@@ -382,7 +383,7 @@ def _clean_up_empty_fields(data_dict, check_field="group_assignment"):
 
 
 def _mask_by_cell_type(data_dict, cell_types: list[int] | int):
-    if not isinstance(cell_types, list):
+    if not (isinstance(cell_types, list) or isinstance(cell_types, omegaconf.listconfig.ListConfig)):
```
thomasZen: Maybe it's easier to do:
```
if isinstance(cell_types, int):
    cell_types = [cell_types]
```
and get rid of the other checks, and then just let np.isin fail.

openretina/utils/file_utils.py
```diff
@@ -26,48 +26,56 @@ def optionally_download_from_url(
 ) -> Path:
     if cache_folder is None:
         cache_folder = get_cache_directory()
+
+    cache_folder = Path(cache_folder)
     download_file_name = Path(*Path(path).parts[-max_num_directories_in_cache_folder:])
+
+    # Search for the file in any subdirectory of cache_folder
+    existing_files = list(cache_folder.rglob(download_file_name.name))
```
thomasZen, line 34: For understanding, we os.exists did not work when downloading folders or was it another problem (dominic also mentioned sth. related I think)?

thomasZen (approved):
added smaller comments though (I can try the data.root_dir stuff in a separate PR)

openretina/eval/metrics.py
fededagos: it is actually fraction of explainable variance explained, which is way I changed it to feve :)

PR #99, opened 2025-02-06
Fix MDS notebook when running on GPU

thomasZen:
I forgot some `.cpu()` calls.
I also removed the hydra imports as we do not need them anymore.


fededagos (approved):
LGTM

PR #104, opened 2025-02-13
Dev: Plotting improvements

thomasZen:
- Use black rectangle around each spatial component (see bottom left plot in attached picture)
- Show responses to MEI and MDS with dotted/straight lines in single plot in MDS notebook

Figure_1


fededagos (approved):
LGTM, can merge if you want

PR #106, opened 2025-02-14
Dev: test for openretina 2025 paper listings

thomasZen:
I added so far the initial example, the mei and the weight visualization listings.


fededagos (approved):
LGTM

PR #109, opened 2025-02-21
Temporal gaussian low pass filter for stimulus optimization

thomasZen:
- Added low pass filter
- Updated MDS notebook to use low pass filter
- Cleaned output of mds notebook
- Fixed LAPLACE_1D kernel (see [Wikipedia Laplace Operator](https://en.wikipedia.org/wiki/Discrete_Laplace_operator#Implementation_via_operator_discretization))
- Wrote simple test for stimulus optimization


openretina/insilico/stimulus_optimization/regularizer.py
```diff
@@ -83,3 +93,36 @@ def process(self, x: torch.Tensor) -> torch.Tensor:
...
+    This acts as a low pass filter."""
+
+    def __init__(
+        self,
+        sigma: float,
+        kernel_size: int,
+        device: str = "cpu",
+    ):
+        kernel = _gaussian_1d_kernel(sigma, kernel_size)
+        self._kernel = kernel.unsqueeze(0).unsqueeze(0).unsqueeze(-1).unsqueeze(-1).to(device)
```
fededagos: I would maybe use einops rearrange here to make it a bit more readable. Alternatively, maybe you could leave the kernel as is (without adding dummy dimensions) and then using einops repeat in the process function which also supports adding dummy dimensions while repeating

openretina/insilico/stimulus_optimization/optimizer.py
```diff
@@ -9,14 +9,25 @@
     StimulusRegularizationLoss,
 )
 
+T = TypeVar("T")
```
fededagos, line 12: Why this? I am just curious

fededagos (approved):
LGTM, minor comments

openretina/insilico/stimulus_optimization/optimizer.py
thomasZen: It's later on used in the type annotation:
```
def convert_to_list(x: list[T] | T | None) -> list[T]:
```

I think that's how you define Generics in python, so that the type checker understands that it works e.g. both if you input a `str` or an `int` above. Also see: https://docs.python.org/3/library/typing.html#generics

openretina/insilico/stimulus_optimization/regularizer.py
thomasZen: Good point, updated it (I have to use einops a bit more often to realize myself when it's useful :D )

PR #110, opened 2025-02-21
Add cmd option to create artificial data and an h5 dataloader

thomasZen:
## Command line tool to create artificial data
```
openretina create-data --help
openretina create-data ./test_data
```
This creates a folder structure like the following:
```
tree ./test_data
test_data
├── session_0.hdf5
├── session_1.hdf5
└── stimuli
    ├── random_noise_0.npy
    ├── random_noise_1.npy
    └── random_noise_2.npy
```

Each h5 file has the following keys (you can also add the stimuli directly in the h5 files which will have priority over loading from the stimuli folder):
```
['responses_random_noise_0', 'responses_random_noise_1', 'responses_random_noise_2', 'session_info']
```

## Training with h5 dataloader
```
openretina create-data ./test_data; \
openretina train --config-path configs --config-name h5_core_readout \
	data.root_dir="." \
```

This will yield the following results:
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃        Test metric        ┃       DataLoader 0        ┃       DataLoader 1        ┃       DataLoader 2        ┃       DataLoader 3        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     test_correlation      │    0.2740490734577179     │    0.2746186852455139     │     0.286388063513597     │    0.28307022257677694    │
│         test_loss         │      -7528.41796875       │     -4413.47607421875     │    -354.35970657182247    │    -350.1444809085195     │
└───────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────┘

```

## Various
- Converted dataclasses into normal classes as hydra complained that types like `Float | dict[str, Float]` are not allowed. I call post_init explicitly now.
- Moved openretina.cli.openretina to openretina.cli.openretina_main as python otherwise gets confused when executing `import openretina...` from the cli directory (it think openretina refers to openretina.cli.openretina then)
- Added simple file_utils tests


openretina/data_io/karamanlis_2024/responses.py
```diff
@@ -87,7 +87,7 @@ def load_all_responses(
         "https://huggingface.co/datasets/open-retina/open-retina/tree/main/gollisch_lab/karamanlis_2024/sessions"
         stim_type (str): The stimulus type to filter files.
         specie (str): Animal species (e.g., "mouse", "marmoset").
-        fr_normalisation (float): Normalization factor for firing rates.
+        fr_normalization (float): Normalization factor for firing rates.
```
thomasZen, line 90: inconsistencies in spelling within line: normalisation vs normalization

openretina/data_io/base.py
```diff
@@ -31,6 +55,16 @@ def __post_init__(self):
...
 
+    @property
+    def test(self) -> np.ndarray:
+        if len(self._test) > 1:
+            raise ValueError(f"Multiple test responses present: {list(self._test.keys())}")
+        return self._test[next(iter(self._test.keys()))]
+
+    @property
+    def test_dict(self) -> dict[str, np.ndarray]:
+        return self._test
```
thomasZen: For backwards compatibility I made the `self._test` private and exposed the dict and the test as properties.

openretina/data_io/base_dataloader.py
```diff
@@ -264,28 +265,23 @@ def handle_missing_start_indices(
...
+    elif split in {"validation", "val"}:
+        assert scene_length is not None, "Scene length or start indices must be provided for validation."
+        interval = scene_length
+    elif split == "test":
+        interval = movie_length
+    else:
+        raise NotImplementedError("Start indices could not be recovered.")
 
-    interval = get_chunking_interval(split)
     return np.arange(0, movie_length, interval).tolist()  # type: ignore
```
thomasZen, line 279: I simplified that slightly, it still leads to the same result as `np.arange(0, movie_length, movie_length).tolist() == [0]`.

openretina/data_io/base_dataloader.py
```diff
@@ -562,5 +559,17 @@ def multiple_movies_dataloaders(
                 scene_length=clip_length,
                 allow_over_boundaries=allow_over_boundaries,
             )
+        # test movies
+        for name in all_movies["test_dict"].keys():
+            movie = all_movies["test_dict"][name]
+            dataloaders[name][session_key] = get_movie_dataloader(
+                movie=movie,
+                responses=neuron_data.response_dict_test[name],
+                split="test",
```
thomasZen, line 568: I explicitly call the split "test" now for all test movies. That fixed the bug I talked about before with this PR (cc: @fededagos )

configs/data_io/hdf5_reader.yaml
```diff
@@ -0,0 +1,30 @@
...
+test_names: 
+  - "random_noise_2"
+  - "random_noise_3"
+
+
+stimuli:
+  _target_: openretina.data_io.h5_dataset_reader.load_stimuli
+  _convert_: object
+  base_data_path: ${data.data_dir}
+  mean_std:
```
fededagos: I think it would be a bit cleaner to separate them in two entries, and maybe call them "norm_mean" and "norm_std"

configs/data_io/hdf5_reader.yaml
```diff
@@ -0,0 +1,30 @@
+# Original videos are 600x800 and need to be downsampled to train models
```
fededagos: Leftover from cloning karamanlis config?

tests/utils/test_file_utils.py
fededagos, line 1: Great to have these

configs/hdf5_core_readout.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - _self_ # values in this config will overwrite the defaults
+
+exp_name: h5_core_readout
+seed: 42
+check_stimuli_responses_match: true
+
+data:
+  root_dir: null # Choose the location of the root directory for the project
+  data_dir: "test_data"
+  output_dir: ${data.root_dir}/openretina_assets/runs
```
fededagos:
```suggestion
  output_dir: ${hydra:runtime.output_dir} # Modify in the "hydra/default.yaml" config
```
To match other outer configs

openretina/data_io/h5_dataset_reader.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class TrainTestStimuliProcessor:
+    """Manages the processing of stimuli. Useful for caching results"""
+
+    def __init__(
+        self,
+        test_stimuli: Iterable[str],
+        name_to_stimulus: dict[str, np.ndarray],
+        mean_std: tuple[float, float] | None,
```
fededagos: Would divide and rename them also here

openretina/data_io/h5_dataset_reader.py
```diff
@@ -0,0 +1,167 @@
...
+            f"provided stimulus size: {stimulus.shape=}, provided_size: {stimulus_size=}"
+        )
+    return True
+
+
+def load_stimuli(
+    base_data_path: str,
+    test_names: Iterable[str],
+    stimulus_size: list[int],
+    mean_std: tuple[float, float] | None,
```
fededagos: Again I would divide and rename here

openretina/cli/create_data.py
fededagos, line 1: Nice, I think having the responses computed from the stimuli is also a nice touch, synthetic data like this can also be useful for us for benchmarking and testing new architectures :) I really like it

openretina/cli/train.py
```diff
@@ -72,13 +72,14 @@ def train_model(cfg: DictConfig) -> None:
...
     log.info(f"Instantiating trainer <{cfg.trainer._target_}>")
-    trainer = hydra.utils.instantiate(cfg.trainer, logger=logger_array, callbacks=callbacks)
+    trainer: lightning.Trainer = hydra.utils.instantiate(cfg.trainer, logger=logger_array, callbacks=callbacks)
     trainer.fit(model=model, train_dataloaders=train_loader, val_dataloaders=valid_loader)
 
     ### Testing
     log.info("Starting testing!")
-    test_loader = ShortCycler(dataloaders["test"])
-    trainer.test(model, dataloaders=[train_loader, valid_loader, test_loader], ckpt_path="best")
+    short_cyclers = [(n, ShortCycler(dl)) for n, dl in dataloaders.items()]
```
fededagos, line 83: I would maybe print the correspondence between dataloader index and dataloader name, so it is clear which one is which and does not need to be inferred.

openretina/data_io/base.py
fededagos, line 1: I have thought about this quite a bit, and I think I would like both to keep being dataclasses. Perhaps we could be more explicit and have both test and test_dict defined in the constructor (throwing an error in post_init if both are present, or if "test" is not an individual array). Then, if only test is present, a default test_dict can still be set in post_init. That way, test_dict will always be there (and can be relied upon), and test will only be there if there is a single test element (and again can be relied upon when coding up functions for this specific case). Thoughts on this? To me it sounds a bit more explicit and direct, but I could be missing something that will not work, as you have digged a bit more into this

openretina/data_io/cyclers.py
```diff
@@ -64,6 +64,7 @@ class ShortCycler(torch.utils.data.IterableDataset):
 
     def __init__(self, loaders: dict[str, DataLoader]):
         self.loaders = loaders
+        self.name = "Foobar"
```
fededagos: What's up here?

fededagos (approved):
Looks good overall! I have suggested some small changes
I would also probably like to keep the movies and responses traintestsplit objects as dataclasses, but if it is too painful it is fine (see if my comment on the PR makes sense).

configs/data_io/hdf5_reader.yaml
thomasZen: I changed that

configs/data_io/hdf5_reader.yaml
thomasZen: Yes, deleted it.

configs/hdf5_core_readout.yaml
thomasZen: Good point, changed.

openretina/data_io/cyclers.py
thomasZen: I wanted to see if pytorch lightning picks up this name, but it did not. I deleted it now :)

openretina/data_io/base.py
thomasZen: I switched back to using dataclasses. Thesethings are imperfect:
```
    # test_dict: dict[str, Float[np.ndarray, "channels train_time height width"]]
    # (dataclass complains about unsupported value type when using full annotation)
    test_dict: dict = field(default_factory=lambda: {})
```
When using the full type annotatio `test_dict: dict[str, Float[np.ndarray, "channels train_time height width"]]`the dataclass throws the following error during initialization:
```
omegaconf.errors.ValidationError: Unsupported value type: 'jaxtyping.Float[ndarray, 'channels train_time height width']'
    full_key: test_dict
    object_type=None
```

I added a function test_movie that returns a single test_movie (also did the same for test responses) or throws an error if there are multiple. It was not possible to cleanly overwrite the `.test` call with `@property`, see https://florimond.dev/en/posts/2018/10/reconciling-dataclasses-and-properties-in-python for a longer explanation.

PR #112, opened 2025-02-21
Change colour maps for response tuning maps

lhoefling:
Change colour maps for response tuning maps, Fig 3 d


thomasZen (approved):
Looks good, I like the new color map.

I think you still have to run `make fix-all` to make the codestyle test work.
We also recently cleared the output of the notebooks before merging. That would get rid of warnings like the following in the notebook:
```
     "text": [
      "/home/bethge/bkr578/.local/lib/python3.11/site-packages/torch/utils/_pytree.py:185: FutureWarning: optree is installed but the version is too old to support PyTorch Dynamo in C++ pytree. C++ pytree support is disabled. Please consider upgrading optree using `python3 -m pip install --upgrade 'optree>=0.13.0'`.\n",
      "  warnings.warn(\n"
     ]
    }
```

thomasZen:
^ I went ahead and did these two things (plus a bit of codestyle in the notebook).

PR #114, opened 2025-02-27
Container definition files

fededagos:
Just added jupyterlab in dev requirements since it is something quite common that people might want to run inside the container, so it also gets installed.


thomasZen (approved):
Looks good (i adapted the dockerfile slightly to make it build)

PR #115, opened 2025-03-03
Automatize Publishing on Pypi

thomasZen:
Todo: Upgrade version to 1.0

Will push to testpypi everytime the version is upgraded.
Will push to pypi when a tag that starts with `release` is pushed.

I followed https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/  for the setup.

This successfully pushed a version to __test__ .pypi: https://test.pypi.org/project/openretina/


fededagos:
Seems to work fine! We might want to fix the readme as well before releasing, and making sure the image is fetched from somewhere that is findable also by pypi when it tries to render.

fededagos (approved):
LGTM

PR #117, opened 2025-03-06
Setting default cache directory at module import

fededagos:
The cache directory is now set to the default location at every module import.
The default root directory in the configs looks up the environment variable by default instead of null.


openretina/sitecustomize.py
fededagos, line 1: This is quite neat, I had no idea it existed:
https://medium.com/alan/simplifying-python-workflows-with-the-sitecustomize-py-e1b1ad5c6fbe

configs/maheswaranathan_2023_LNP.yaml
```diff
@@ -19,7 +19,7 @@ seed: 42
 check_stimuli_responses_match: false
 
 data:
-  root_dir: ???
+  root_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY}
   data_dir: ${data.root_dir}/baccus_data/neural_code_data/ganglion_cell_data/
   output_dir: ${data.root_dir}/openretina_assets/runs
```
thomasZen:
```suggestion
  ${hydra:runtime.output_dir}
```
for consistency with the other files.

configs/template_outer_config.yaml
```diff
@@ -18,7 +18,7 @@ seed: 42
 check_stimuli_responses_match: false
 
 data:
-  root_dir: null # Choose the location of the root directory for the project
+  root_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY} # Choose the location of the root directory for the project
   data_dir: ${data.root_dir}/... # Choose the location of the data. Should be used in data_io functions.
   output_dir: ${data.root_dir}/openretina_assets/runs
```
thomasZen:
```suggestion
  output_dir: ${hydra:runtime.output_dir}
```
for consistency with the other files

openretina/cli/train.py
```diff
@@ -29,6 +29,9 @@ def train_model(cfg: DictConfig) -> None:
     ### Set cache folder
     os.environ["OPENRETINA_CACHE_DIRECTORY"] = cfg.data.root_dir
 
+    ### Display log directory for ease of access
+    log.info(f"Saving run logs at: {cfg.data.output_dir}")
```
thomasZen: I'm not sure if cfg.data.output_dir is a perfect path, something like cfg.training.output_dir would in theory be better (but probably does not make sense to change that now)

configs/karamanlis_2024_core_readout.yaml
```diff
@@ -18,7 +18,7 @@ seed: 42
 check_stimuli_responses_match: true
 
 data:
-  root_dir: null # Choose the location of the root directory for the project
+  root_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY} # Choose the location of the root directory for the project
   data_dir: https://huggingface.co/datasets/open-retina/open-retina/tree/main/gollisch_lab/karamanlis_2024/sessions
```
thomasZen, line 23: When do we specify the url here? Above we put a path but specify movies path and responses path with url.
I can see how this happened, but might be a bit confusing.

configs/maheswaranathan_2023_LNP.yaml
```diff
@@ -19,7 +19,7 @@ seed: 42
 check_stimuli_responses_match: false
 
 data:
-  root_dir: ???
+  root_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY}
   data_dir: ${data.root_dir}/baccus_data/neural_code_data/ganglion_cell_data/
```
thomasZen:
```suggestion
  data_dir: https://huggingface.co/datasets/open-retina/open-retina/resolve/main/baccus_lab/maheswaranathan_2023/neural_code_data.zip
```
for consistency with the core_readout setup below.

openretina/__init__.py
```diff
@@ -0,0 +1,5 @@
+import os
+
+from openretina.utils.file_utils import get_cache_directory
```
thomasZen: This ends up importing these external packages from file_utils:
```
import requests
import tenacity
from huggingface_hub import HfApi, hf_hub_download
from huggingface_hub.errors import EntryNotFoundError
from tqdm.auto import tqdm
```
that's okay I think, but just to keep in mind that we do not endlessly add things. E.g I would not want to import tensorflow or torch everytime I import openretina.

configs/hoefling_2024_core_readout_high_res.yaml
```diff
@@ -34,7 +34,7 @@ model:
   # downsample_input_kernel_size: [1, 4, 4]  # to reproduce 18x16px lowres input
 
 data:
-  root_dir: null # Choose the location of the root directory for the project  
+  root_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY} # Choose the location of the root directory for the project  
   data_dir: ${data.root_dir}/data/ # Modify according to preferred data location
```
thomasZen: The movies and responses are downloaded to this folder, right?
If that's true, we should either change it to `${oc.env:OPENRETINA_CACHE_DIRECTORY}` or adapt the notebooks which currently use:
```
data_path = get_local_file_path(
    "https://huggingface.co/datasets/open-retina/open-retina/blob/main/baccus_lab/maheswaranathan_2023/neural_code_data.zip"
)
```
and therefore default to the default cache directory, we can set cache_folder explicitly there.

I think at least for larger datasets it's relevant that the cache directory is hit and we do not download it twice at different locations.

configs/hdf5_core_readout.yaml
```diff
@@ -18,7 +18,7 @@ seed: 42
 check_stimuli_responses_match: true
 
 data:
-  root_dir: null # Choose the location of the root directory for the project
+  root_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY} # Choose the location of the root directory for the project
```
thomasZen: This is somewhat circular. In train.py we do:
```
    ### Set cache folder
    os.environ["OPENRETINA_CACHE_DIRECTORY"] = cfg.data.root_dir
```

thomasZen (commented):
I think way we set the path is a bit convoluted and sometimes circular now. Here's one suggestion:

Decouple the cache folder from the root_dir:
- The cache folder should be used to store data and finalized (= completely trained) models that are downloaded from huggingface/gin.
- The root dir is the directory where you store training artefacts likes logs, intermediate checkpoints, etc. Maybe we can set it to `.` on default?

This also has the advantage that we do not "trash" the cache folder with more and more data in the Github CI. If I understand it correctly, currently we add the a new trained checkpoint and logs to the cache folder on every test run, and this probably is kept in the cache folder in the github CI.

openretina/__init__.py
fededagos: That is a good point actually... Would it be better to keep it simpler and maybe duplicate only that function? Or put it in another module? Or is it fine like this?

configs/hdf5_core_readout.yaml
fededagos: I will test if that can be removed now, but I think it would need to stay. If a user changes the root_dir it does not change immediately the cache directory, so it would need to be set for functions that do not use it explicitly. If they set the cache directory with an export, then it just sets it again, but it is true that it is a bit circular... Will think about what to do here

thomasZen (approved):
LGTM

PR #120, opened 2025-03-13
Dev: simplify multiple_movies_dataloaders method

### LINKED ISSUE #79: Refactor get_movie_splits in base_dataloader
I think this should not be a function.
It's only used once and just putting it inline or letting the generate_movie_function directly return this dictionary (or an equivalent object) seems better.

_Originally posted by @thomasZen in https://github.com/open-retina/open-retina/pull/77#discussion_r1873381472_

thomasZen:
Fixes https://github.com/open-retina/open-retina/issues/79 and shortens some code.


fededagos (approved):
Looks good!

PR #125, opened 2025-04-10
Update LICENSE

fededagos:
GPL is not permissive enough: repositories under MIT could not easily re-use code from open-retina.


thomasZen (approved):
Looks good, especially after the feedback we got at Cosyne

PR #128, opened 2025-05-08
Move requirements into pyproject, add standard-imghdr

thomasZen:
Additionally run mypy checks on python 3.10 until 3.13.

Standard-imghdr was removed from newer versions of python, and apparently figures out the video format of some files:
> The [imghdr](https://docs.python.org/3.12/library/imghdr.html#module-imghdr) module is a simple tool to guess the image file format from the first 32 bytes of a file or buffer. It supports only a limited number of formats and neither returns resolution nor color depth.
There was a report that the openretina demo notebook did not run without installing it. I could not reproduce it, but I think it's reasonable to add this small dependency.

For some reason I did not figure out how to move the main requirements into pyproject.yaml. When I did pip was not able to resolve the metadata of the packages anymore. I spent half an hour on this but did not figure out why.


fededagos (approved):
Everything looks good. I just created #129 to make sure that we still have at the back of our mind to remove requirements.txt at some point, but I agree it is not too high priority. It is already good to have removed the extra requirements files.

PR #130, opened 2025-05-15
Dev/baptiste/hparams search

BaptisteLorenzi:
First pass on hyperparams search with an example. I switched from my data to Hoefling 2024 low res without any issues.

MLFlow needs to be added to the test environment. As expected there are also some type checks mistake due to the new behavior of the train function (send a float and not None in multirun).

I also renamed the tutorials but git handles notebooks poorly so maybe ignore those for now. In that case, just add 'tuto_4_...'

I also believe the .hydra folder can be ignored. i wasn't sure whether to add it to .gitignore or not.


openretina/cli/train.py
```diff
@@ -10,16 +10,29 @@
...
+@hydra.main(
+    version_base="1.3",
+    config_path="../../configs",
+    config_name="hoefling_2024_core_readout_high_res",
+)
+def main(cfg: DictConfig) -> float | None:
+    score = train_model(cfg)
+    if cfg.objective_target is not None:
+        return score
+    return None
```
thomasZen: Do you need this if? Why not just always return the score?
```suggestion
    return score
```

openretina/cli/train.py
```diff
@@ -10,16 +10,29 @@
...
 
 
-def train_model(cfg: DictConfig) -> None:
+def train_model(cfg: DictConfig) -> float | None:  ## normally train_model
+    # Get the current trial number
+    # Update the trial number in the configuration
```
thomasZen: maybe delete the comments or move them to the location where that happens.
If you want to add a longer description what the function does it's best to use docstrings like:
```suggestion
def train_model(cfg: DictConfig) -> float | None:
   """ Trains a model and returns a score in case the xyz config parameter was set or so """
```

openretina/cli/train.py
```diff
@@ -87,6 +102,50 @@ def train_model(cfg: DictConfig) -> None:
...
+                        "lightning_version": lightning.__version__,
+                    }
+                    mlflow.log_dict(sys_info, "system_info.json")
+
+                    # Log model summary if available
+                    if hasattr(model, "summarize"):
+                        mlflow.log_text(str(model.summarize()), "model_summary.txt")
+            except Exception as e:
+                log.warning(f"Failed to log artifacts to MLflow: {str(e)}")
+            break
```
thomasZen: I think the logic looks good.

It's normally easier to read if there a less indentation levels. Maybe something like this would work, too:
```
mlflow_logger_array = [logger for logger in logger_array if "mlflow" in str(type(logger)).lower()]
if len(mlflow_logger_array > 1:
   raise ValueError(f"Multiple mlflow loggers defined:  {[str(type(logger) for logger in mlflow_logger_array]}")
elif if len(mlflow_logger_array == 1:
    log_to_mlflow(logger, model, ...)
```
and then put the rest of the logic in the log_to_mlflow function

openretina/cli/train.py
```diff
@@ -87,6 +102,50 @@ def train_model(cfg: DictConfig) -> None:
...
+                        "platform": platform.platform(),
+                        "lightning_version": lightning.__version__,
+                    }
+                    mlflow.log_dict(sys_info, "system_info.json")
+
+                    # Log model summary if available
+                    if hasattr(model, "summarize"):
+                        mlflow.log_text(str(model.summarize()), "model_summary.txt")
+            except Exception as e:
+                log.warning(f"Failed to log artifacts to MLflow: {str(e)}")
```
thomasZen:
```suggestion
                log.exception(f"Failed to log artifacts to MLflow: {str(e)}")
```
If you use log.exception it will log the whole stack trace. Not sure if you want this here, but it's useful to know

openretina/cli/train.py
```diff
@@ -87,6 +102,50 @@ def train_model(cfg: DictConfig) -> None:
...
+            break
+
+    if cfg.objective_target is not None:
+        ### Final validation for optuna
+        log.info("Starting validation for Optuna")
+        target_score = trainer.validate(model, dataloaders=[valid_loader], ckpt_path="best")[0][cfg.objective_target]
+        if target_score is None:
+            log.error(f"Objective target '{cfg.objective_target}' is None!")
```
thomasZen:
```suggestion
            log.error(f"Score for objective target '{cfg.objective_target}' is None!")
```

notebooks/tuto_3_new_datasets_guide.ipynb
```diff
@@ -242,8 +242,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "test_stimulus = test_stimulus[None, ...]\n",
-    "train_stimulus = train_stimulus[None, ...]\n",
+    "test_stimulus = test_stimulus.reshape(1, -1, test_stimulus.shape[1], test_stimulus.shape[2])\n",
+    "train_stimulus = train_stimulus.reshape(1, -1, train_stimulus.shape[1], train_stimulus.shape[2])\n",
```
thomasZen: did this lead to an error before?

thomasZen:
> I also renamed the tutorials but git handles notebooks poorly so maybe ignore those for now. In that case, just add 'tuto_4_...'

I agree that git handles notebooks purely :D So far we mostly commited the cleared notebooks without any outputs, but they get ran in the end-to-end tests so we are sure they work. Any specific reason to switch to numbered tutorials (as in tutorial_4 instead of something like hyperparmeter_tuning.ipynb)? 


> I also believe the .hydra folder can be ignored. i wasn't sure whether to add it to .gitignore or not.

It might be good to include an example config in the configs folder, maybe we can add it to the configs/hydra folder in a useful way (ie here: https://github.com/open-retina/open-retina/tree/main/configs/hydra).
Are hydra configs normally stored in/written to `.hydra`? If it's automatically written I'd put it into .gitignore though.

notebooks/tuto_4_hyperparameter_search_hoefling_2024.ipynb
```diff
@@ -0,0 +1,175 @@
...
+   "source": [
+    "# Launching the search\n",
+    "\n",
+    "We will use the same command as usual, adding the option --multirun.\n",
+    "\n",
+    "```bash\n",
+    "openretina train --config-name \"hoefling_2024_core_readout_low_res_hyperparams_search\" --multirun\n",
+    "```\n",
+    "\n",
+    "If the option --multirun is not selected, the sweeper section will be ignore, launching a single training session with the default parameters.\n"
```
thomasZen: I think we can directly add the sweeper section to the example config then.

notebooks/tuto_4_hyperparameter_search_hoefling_2024.ipynb
```diff
@@ -0,0 +1,175 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "0b388760",
+   "metadata": {},
+   "source": [
+    "In this notebook we will go a bit further in the training of a Core/Readout model. Using once again data from Hoefling et al., 2024: [\"A chromatic feature detector in the retina signals visual context changes\"](https://elifesciences.org/articles/86860).\n",
+    "\n",
+    "We will see how we can search the hyperparameter space using Hydra, while doing minimal changes to the configuration files. It can only be used using the CLI version of open_retina."
```
thomasZen: maybe additionally mention and link to Optuna here directly. Later on I thought where do I best find out which optuna keywords there are, so already having a link to optuna handy would get rid of the need to google it.

thomasZen (commented):
Thanks for adding this. I only read the code so far and it looks good, will also try out the notebook later!

openretina/cli/train.py
BaptisteLorenzi: This is just an old comment I forgot to delete, now deleted

openretina/cli/train.py
BaptisteLorenzi: Okay, I'm now using you're suggested test. As for the logic I moved it to an .utls file.

openretina/cli/train.py
BaptisteLorenzi: I don't think we want the whole trace since I wouldn't consider MLFlow a priority in the log. But good to know for sure.

openretina/cli/train.py
BaptisteLorenzi: Sure, score is already None in there is no objective target, as handled in train_model().

notebooks/tuto_4_hyperparameter_search_hoefling_2024.ipynb
BaptisteLorenzi: Why not, but the config has to feature some lines that will be useless (hydra,sweeper), wouldn't that be too confusing ?

BaptisteLorenzi:
> > I also renamed the tutorials but git handles notebooks poorly so maybe ignore those for now. In that case, just add 'tuto_4_...'
> 
> I agree that git handles notebooks purely :D So far we mostly commited the cleared notebooks without any outputs, but they get ran in the end-to-end tests so we are sure they work. Any specific reason to switch to numbered tutorials (as in tutorial_4 instead of something like hyperparmeter_tuning.ipynb)?
> 
> > I also believe the .hydra folder can be ignored. i wasn't sure whether to add it to .gitignore or not.
> 
> It might be good to include an example config in the configs folder, maybe we can add it to the configs/hydra folder in a useful way (ie here: https://github.com/open-retina/open-retina/tree/main/configs/hydra). Are hydra configs normally stored in/written to `.hydra`? If it's automatically written I'd put it into .gitignore though.

> 1. That's also how I handle notebook. The name thing was a minor change, I took it away.
>2. I'v already added a config example hoefling_2024_core_readout_low_res_hyperparams_search . I don't recommend merging it with the initial config. Even though it would be possible since the sweeper is not called when --multirun is not activated, it would make the config file harder to read for new users.

notebooks/tuto_3_new_datasets_guide.ipynb
BaptisteLorenzi: I must have changed those lines by mistakes since there are different on the main branch. Corrected.

.gitignore
```diff
@@ -180,3 +180,9 @@ model-cache/
 openretina_cache_folder/
 openretina_assets/
 test_data/
+mlflow/
+multirun/
+notebooks/lightning_logs
+notebooks/notebook_example/tensorboard
+checkpoints/
+tensorboard/
```
thomasZen, line 188:
```suggestion
tensorboard/

```

Nitpick: It's the Posix standard to end files in a new line. (if you are interested why, see here for an explanation: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline :) )

configs/logger/mlflow.yaml
```diff
@@ -0,0 +1,9 @@
+# MLflow logger configuration
+mlflow:
+  _target_: lightning.pytorch.loggers.MLFlowLogger
+  experiment_name: ${exp_name}
+  tracking_uri: "${paths.log_dir}/openretina_assets/mlflow"
+  log_model: false # Handled manually in the training script
+  run_name: ${exp_name}_run_${now:%Y%m%d_%H%M%S}
+  tags:
+    user: "${oc.env:USER,unknown}"
```
thomasZen:
```suggestion
    user: "${oc.env:USER,unknown}"

```

notebooks/openretina_demo.ipynb
```diff
@@ -436,7 +436,7 @@
     "# Display the widgets\n",
     "display(widgets.VBox([video_dropdown, loading, video_output, empty], layout=box_layout))\n",
     "\n",
-    "# Initial video display\n",
+    "# Initial video displaya\n",
```
thomasZen:
```suggestion
    "# Initial video display\n",
```
this comment seems to be off.

.hydra/config.yaml
```diff
@@ -0,0 +1,128 @@
...
+    base_data_path: ${paths.data_dir}
+    stim_type: flashednaturalscene
+    normalize_stimuli: true
+  responses:
+    _target_: openretina.data_io.lorenzi_2025.responses.load_all_responses
+    _convert_: object
+    base_data_path: ${paths.data_dir}
+    stim_type: flashednaturalscene
+    response_type: firing_rate_25ms
+    fr_normalization: 1.0
```
thomasZen: Note: this points to functions that do not yet exists

.hydra/hydra.yaml
```diff
@@ -0,0 +1,178 @@
...
+    - path: hydra.conf
+      schema: pkg
+      provider: hydra
+    - path: /home/baptiste/Documents/LabPipelines/open-retina/configs
+      schema: file
+      provider: main
+    - path: ''
+      schema: structured
+      provider: schema
+    output_dir: /home/baptiste/Documents/LabPipelines/open-retina
```
thomasZen: also here are paths like `/home/baptiste/Documents`. Maybe change path to `TODO` or `/path/to/open-retina` or set it to `.` and `./configs` or so.

openretina/cli/train.py
```diff
@@ -10,16 +10,22 @@
...
+    config_path="../../configs",
+    config_name="hoefling_2024_core_readout_high_res",
+)
+def main(cfg: DictConfig) -> float | None:
+    score = train_model(cfg)
+    return score
 
 
-def train_model(cfg: DictConfig) -> None:
+def train_model(cfg: DictConfig) -> float | None:  ## normally train_model
```
thomasZen:
```suggestion
def train_model(cfg: DictConfig) -> float | None: 
```

I'd remove the comment here as it doesn't add information I think

notebooks/tuto_4_hyperparameter_search_hoefling_2024.ipynb
thomasZen: ok, then let's keep it as is.

thomasZen (approved):
Looks good, added a few more small comments.
Feel free to merge afterwards.

notebooks/hyperparameter_search_hoefling_2024.ipynb
```diff
@@ -0,0 +1,207 @@
+{
```
thomasZen, line 1: <p>Optional. Enable loggers. (as it's not only the mlflow logger)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/open-retina/open-retina/pull/130/discussion/#comment-2120351092'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

thomasZen, line 1: <p>small typo: be ignore &gt; be ignored.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/open-retina/open-retina/pull/130/discussion/#comment-2120351096'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

.hydra/config.yaml
BaptisteLorenzi: The entire .hydra folder has just nothing to do on git, I planned to put it on gitignore before but it can simply be removed. It must have been crated automatically be the pipeline while I was testing other code.

.hydra/hydra.yaml
BaptisteLorenzi: Same as above

PR #132, opened 2025-05-27
Training readout only

SamuelSuhai:
I added three things relating to fine-tuning models or training models on new data: 
1) An `only_train_readout`  config parameter for `train.py`. With this you can freeze the core and only train the readout layers. 
2) With the `load_model_path` config parameter you can load a model and train it on new data. Some data related attributes are then updated in with a `update_model_data_info` in `BaseCoreReadout`. 
3) I added a add_sessions method to `MultiGaussianReadoutWrapper` to add sessions to already instantiated models.
Based on these changes you could then load a model trained on some data and then continue training it (or only training the readout) based on data from a new session.


SamuelSuhai:
If you want to test the freezing core parameters functionality you can go to commit `[8b3f571](https://github.com/open-retina/open-retina/pull/132/commits/8b3f571f66d63481253c50427df5aecd5ec62f7c)` there I still have a log_parameter_snapshots function in `train.py`  which logs whether or not core or readout parameters changed. To test the code, I trained the hoefling_2024_core_readout_low_res model, added a dummy dataset that simply has different session names (but identical data), loaded the trained model, added a new readout, and only trained the new readout on "dummy" dataset. I made sure that the core does not change and looked at the new readout parameters compared to the readout parameters from the previous training run. As one would expect the readout parameters are very similar but not identical. Here some plots if interested: 
[readout_param_correl_mat.zip](https://github.com/user-attachments/files/20454856/readout_param_correl_mat.zip)

configs/hoefling_2024_core_readout_high_res.yaml
```diff
@@ -51,3 +52,5 @@ trainer:
   accumulate_grad_batches: 10
   gradient_clip_val: 1
   precision: 16-mixed
+
+only_train_readout: false # whether or not to freeze core parameters and only train readout (eg. when finetuning model to new data)
```
fededagos: Maybe add an empty line at the end of the document :)

```suggestion
only_train_readout: false # whether or not to freeze core parameters and only train readout (eg. when finetuning model to new data)

```

configs/hoefling_2024_core_readout_low_res.yaml
```diff
@@ -40,3 +41,5 @@ trainer:
   max_epochs: 100
   accumulate_grad_batches: 1
   gradient_clip_val: 1
+
+only_train_readout: false # whether or not to freeze core parameters and only train readout (eg. when finetuning model to new data)
```
fededagos: Same here :)

```suggestion
only_train_readout: false # whether or not to freeze core parameters and only train readout (eg. when finetuning model to new data)

```

fededagos (approved):
Looks good!!

This is suuper minor, but I flagged two missing new lines at the end of the config files (see [here](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file), Thomas made me aware of this originally :) ).

openretina/modules/readout/multi_readout.py
```diff
@@ -28,27 +28,40 @@ def __init__(
...
+        self.gaussian_masks = gaussian_masks
+        self.readout_reg_avg = readout_reg_avg
+
+    def add_sessions(self, n_neurons_dict: dict[str, int]) -> None:
+        """Adds new sessions to the readout wrapper.
+        Can be called to add new sessions to an existing readout wrapper."""
+
+        assert all(key not in self.keys() for key in n_neurons_dict), (
+            "Found duplicate sessions in n_neurons_dict. Make sure to use different session names for each session."
+        )
```
thomasZen:
```suggestion
        if any(key in self.keys() for key in n_neurons_dict):
            duplicate_session_names = set(self.keys()).intersection(n_neuron_dict.keys())
            raise ValueError(f"Found duplicate sessions in n_neurons_dict:  {duplicate_session_names=}. Make sure to use different session names for each session."
        )
```

Normally you use assert to check the internal consistency within a class. As this is also called from other places, I'd prefer raising an error. The other suggestion is to also print duplicate sessions in the error.

configs/hoefling_2024_core_readout_high_res.yaml
```diff
@@ -34,6 +34,7 @@ model:
   # downsample_input_kernel_size: [1, 4, 4]  # to reproduce 18x16px lowres input
 
 paths:
+  # load_model_path: # e.g. hoefling_2024_base_high_res or path to a .ckpt model
```
thomasZen:
```suggestion
  load_model_path: None  # e.g. hoefling_2024_base_high_res or path to a .ckpt model
```

I'd explicitly set this to None, as then it's not commented out. Also in the config below.

thomasZen (approved):
Looks good.
Added a two small suggestions (feel free to ignore if you don't agree though)

configs/hoefling_2024_core_readout_high_res.yaml
thomasZen: for context why that's considered better (I just looked for it because it came up in another PR too):
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

PR #134, opened 2025-06-09
Support for distributed training.

fededagos:
After doing some [reading on distributed training](https://huggingface.co/spaces/nanotron/ultrascale-playbook), I wanted to try to finally make openretina compatible with distributed training, which is one of the main things PyTorch lightning makes easy. 

I have created a new MAP-style dataloader to replace the cyclers, and added a new distributed training script that uses by default the "ZeRO-2" strategy by deepspeed: it copies the model on different devices, but shards activations, optimiser states and data across them. 

This is the best strategy for us right now as:
1. Vanilla DDP does not provide any memory advantages
2. ZeRO-3 (or FSDP) would require the model to be sharding-compatible, but since not all of the model is used during each forward pass, it would require changing it internally. Also, our models are not huge usually, so we can afford to not shard them especially if activations and optimiser states get distributed.

There is no direct torch implementation of ZeRO-2, which is why I added a dependency on deepspeed.

Note that right now for the script to work it is assumed the user has access to more than 1 GPU in a single interactive node. On SLURM, this still requires starting the node with n-tasks=n_available_gpus.

Starting directly a SLURM script, not in an interactive node, is a bit more complicated and would require a dependency on the submitit plugin for hydra. Perhaps we could do this in the future if we really need large-scale training at some point.
Right now I have added this mostly to put in practice what I was learning, and to make training of the biggest model a bit faster.


pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "jupyter",
     "ipywidgets",
     "mlflow",
+    "deepspeed",
```
thomasZen, line 43: I'm a bit skeptical of putting this in the main requirements, maybe putting them as a separate optional section is a good option (at least if we do not think that distributed training will be our main training workflow)

openretina/cli/distributed_train.py
```diff
@@ -0,0 +1,189 @@
...
+    ]
+
+    ### Trainer init
+    log.info(f"Instantiating trainer <{cfg.trainer._target_}>")
+    trainer: lightning.Trainer = hydra.utils.instantiate(cfg.trainer, logger=logger_array, callbacks=callbacks)
+    trainer.fit(model=model, datamodule=data_module)
+
+    ### Convert DeepSpeed checkpoint to regular PyTorch format if needed
+    is_deepspeed = "deepspeed" in str(trainer.strategy).lower()
+    if is_deepspeed:
```
thomasZen, line 111: Does distributed training also work with another strategy?

openretina/cli/distributed_train.py
```diff
@@ -0,0 +1,189 @@
...
+            checkpoint_patterns = [
+                os.path.join(checkpoint_dir, "epoch=*_val_correlation=*.ckpt"),
+                os.path.join(checkpoint_dir, "epoch*.ckpt"),
+                os.path.join(checkpoint_dir, "*.ckpt"),
+            ]
+
+            latest_checkpoint = None
+            for pattern in checkpoint_patterns:
+                checkpoints = glob.glob(pattern)
+                if checkpoints:
```
thomasZen, line 128: This does seem a bit instable. Was this copied over? If not maybe at least adding an option to explicitly define a checkpoint path might make sense.

openretina/cli/distributed_train.py
```diff
@@ -0,0 +1,189 @@
...
+                output_path = os.path.join(cfg.paths.output_dir, "model_final.pt")
+                log.info(f"Converting DeepSpeed checkpoint from {latest_checkpoint} to {output_path}")
+                convert_zero_checkpoint_to_fp32_state_dict(latest_checkpoint, output_path)
+                log.info(f"Successfully converted DeepSpeed checkpoint to {output_path}")
+            else:
+                log.warning(f"No DeepSpeed checkpoint directory found in {checkpoint_dir}")
+
+        except ImportError:
+            log.warning(
+                "DeepSpeed checkpoint conversion not available - lightning.pytorch.utilities.deepspeed not found"
```
thomasZen, line 144: Maybe say here that you should install some python library to make it work

thomasZen: or  something like `openretina[distributed]` if you put the dependency as optional

openretina/cli/distributed_train.py
```diff
@@ -0,0 +1,189 @@
...
+        ]
+        if target_score is None:
+            log.error(f"Score for objective target '{cfg.objective_target}' is None!")
+        return target_score
+    else:
+        return None
+
+
+if __name__ == "__main__":
+    main()
```
thomasZen, line 189: I think this is fine as it is kind of a non core functionality, but there's a lot of code duplication with train.py.

And maybe call everything `train-distributed` instead of `distributed-train` just for the reason that people mostly order things alphabetically and then these scripts and commands are next to each other.

openretina/cli/openretina_main.py
```diff
@@ -47,15 +47,73 @@ def _train(cfg: DictConfig) -> float | None:
...
     subparsers = parser.add_subparsers(dest="command", help="Available commands")
 
     # Train command
     train_parser = subparsers.add_parser("train", help="Train a model")
     train_parser.add_argument("--config-path", default=None, help="Path to config directory")
 
+    # Distributed train command
+    distributed_train_parser = subparsers.add_parser(
+        "distributed-train", help="Train a model with distributed training"
+    )
```
thomasZen, line 114: Maybe "train-distributed" is a better name as it then starts with the same prefix as the train command.

openretina/data_io/cyclers.py
```diff
@@ -57,14 +57,17 @@ def __len__(self):
...
+class ShortCycler(IterableDataset):
     """
     Cycles through the elements of each dataloader without repeating any element.
     """
 
     def __init__(self, loaders: dict[str, DataLoader]):
         self.loaders = loaders
 
+    def __len__(self):
+        return sum(len(loader) for loader in self.loaders.values())
```
thomasZen, line 69: didn't we remove this for some reason in the past?

openretina/data_io/session_combined_dataset.py
```diff
@@ -0,0 +1,563 @@
...
+
+        # Pre-compute sampling indices for this epoch
+        self._update_sampling_indices()
+
+    def set_epoch(self, epoch: int):
+        """Set the epoch for deterministic sampling."""
+        self._current_epoch = epoch
+        self._update_sampling_indices()
+
+    def _update_sampling_indices(self):
```
thomasZen, line 105:
```suggestion
    def _update_sampling_indices(self) -> None:
```

openretina/data_io/session_combined_dataset.py
```diff
@@ -0,0 +1,563 @@
...
+
+        # Get session lengths
+        self.session_lengths = {key: len(dataset) for key, dataset in session_datasets.items()}
+        self.max_session_length = max(self.session_lengths.values())
+        self.total_session_length = sum(self.session_lengths.values())
+
+        # Build the global index mapping
+        self._build_index_mapping()
+
+    def _build_index_mapping(self):
```
thomasZen, line 178:
```suggestion
    def _build_index_mapping(self) -> None:
```

openretina/data_io/session_combined_dataset.py
```diff
@@ -0,0 +1,563 @@
...
+                # Add all batches from this session before moving to the next
+                for batch_idx in range(num_batches):
+                    # Add batch_size consecutive indices for this batch
+                    for item_in_batch in range(self.batch_size):
+                        # Calculate the actual dataset index for this item in the batch
+                        dataset_idx = batch_idx * self.batch_size + item_in_batch
+                        # If we go beyond the dataset, wrap around (for incomplete last batch)
+                        if dataset_idx >= dataset_length:
+                            dataset_idx = dataset_idx % dataset_length
+                        self.index_mapping.append((session_key, dataset_idx))
```
thomasZen, line 252: else throw error, at least if it's a error if you use a different cycling mode.

openretina/data_io/session_combined_dataset.py
```diff
@@ -0,0 +1,563 @@
...
+
+    def set_epoch(self, epoch: int):
```
thomasZen, line 254:
```suggestion
    def set_epoch(self, epoch: int) -> None:
```

openretina/data_io/session_combined_dataset.py
```diff
@@ -0,0 +1,563 @@
...
+            )
+
+    def train_dataloader(self):
+        return self.train_loader
+
+    def val_dataloader(self):
+        return self.val_loader
+
+    def test_dataloader(self):
+        """Return test dataloaders for all splits (train, validation, test)."""
```
thomasZen, line 541: What's the test_dataloader for the train split?
isn't the function name and docstring a bit inconsistent (test_dataloader vs test_dataloaders, etc)

thomasZen (commented):
I skimmed the code and am happy to approve it after making it an optional dependency, and after we tested that normal training still works.
I think the addition might be very cool for bigger or more computational models, but on the other hand it also a lot of additional and I'm unsure if we'll use it much (but maybe you already have a use case in mind :))

It would also be cool to add an benchmark of the high res model training in the PR description (basically how much speedier it is on 4/8 GPUs compared to on one GPU before this PR, and maybe also mention test correlation there)

Issue #136, opened 2025-07-17
Model "hoefling_2024_base_low_res" is not causal


thomasZen:
The model "hoefling_2024_base_low_res" and probably also the high res model is non causal as it uses the temporal padding in the second convolution, see:
https://github.com/open-retina/open-retina/blob/5455c6fb5aa63e6901dba900b2ab091aca65c3f2/configs/model/GRU_core_readout.yaml#L15

This leads to the model responses reacting to input that is up to 5 frames "in the future" (among other things this can be confusing when optimizing MEIs). We should probably get rid of temporal padding all together for our default models.

thomasZen:
Also see: https://github.com/open-retina/open-retina/pull/84/files#r1884129043

thomasZen:
For training new models this is fixed with https://github.com/open-retina/open-retina/pull/138.
I keep this issue open until we actually uploaded the new models, though.

thomasZen:
This was fixed with the most recent release and the new uploaded models.

PR #138, opened 2025-07-28
Adapt core_readout_padding to make model causal

thomasZen:
See: https://github.com/open-retina/open-retina/issues/136

Also added a causality test.

For Lara's ensemble model the padding was and the ensemble model is causal:
```
>>> from openretina.utils.nnfabrik_model_loading import load_ensemble_model_from_remote                                                                                         
>>> data_info, ensemble_model = load_ensemble_model_from_remote(device="cpu") 
...
>>> ensemble_model.members[0].core.features[0].conv.padding
0
>>> ensemble_model.members[0].core.features[1].conv.padding
(0, 2, 2)
```
I used the same setup, and confirmed that freshly trained models are causal now for the hoefling 2024 setup.

Also all our models are currently none causal, we should retrain and reupload all of them:
```
FAILED tests/models/test_core_readout.py::test_model_is_causaul[hoefling_2024_base_high_res] - assert False
FAILED tests/models/test_core_readout.py::test_model_is_causaul[hoefling_2024_base_low_res] - assert False
FAILED tests/models/test_core_readout.py::test_model_is_causaul[karamanlis_2024_base] - assert False
FAILED tests/models/test_core_readout.py::test_model_is_causaul[maheswaranathan_2023_base] - assert False
```


openretina/utils/debug_model.py
fededagos, line 1: This looks super useful going forward!

fededagos (approved):
Looks good to me! It is great to have a test for this going forward, we can re-train everything after the hackathon, upload all models and then turn it on

PR #139, opened 2025-08-01
Update README.md

lhoefling:
explicitly add links to datasets

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,11 +59,17 @@ The openretina package is structured as follows:
 - utils: Utility functions that are used across above submodules
 
 
-## Related papers
+## Related papers and data sources
 
 The Core + Readout model was developed in the paper [A chromatic feature detector in the retina signals visual context changes](https://elifesciences.org/articles/86860). All datasets used in openretina are shared under a CC-BY Share-Alike license, and we acknowledge and credit the original sources below:
-- hoefling_2024: Originally published by Höfling et al. (2024), eLife: [A chromatic feature detector in the retina signals visual context changes](https://doi.org/10.7554/eLife.86860).
-- karamanlis_2024: Originally published by Karamanlis et al. (2024), Nature: [Nonlinear receptive fields evoke redundant retinal coding of natural scenes](https://doi.org/10.1038/s41586-024-08212-3)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,11 +59,17 @@ The openretina package is structured as follows:
...
+   - Paper: [A chromatic feature detector in the retina signals visual context changes](https://doi.org/10.7554/eLife.86860).
+   - Dataset originally deposited at: https://gin.g-node.org/eulerlab/rgc-natstim
+- karamanlis_2024: Originally published by Karamanlis et al. (2024), Nature.
+  - Paper: [Nonlinear receptive fields evoke redundant retinal coding of natural scenes](https://doi.org/10.1038/s41586-024-08212-3)
+  - Dataset: Karamanlis D, Gollisch T (2023) Dataset - Marmoset and mouse retinal ganglion cell responses to natural stimuli and supporting data. G-Node. https://doi.org/10.12751/g-node.ejk8kx 
+- maheswaranathan_2023: Originally published by Maheswaranathan et al. (2023), Neuron
+  - Paper: [Interpreting the retinal neural code for natural scenes: From computations to neurons](https://doi.org/10.1016/j.neuron.2023.06.007)
+  - Dataset: Maheswaranathan, N., McIntosh, L., Tanaka, H., Grant, S., Kastner, D., Melander, J., Nayebi, A., Brezovec, L., Wang, J. Ganguli, S. Baccus, S. (2023). Interpreting the retinal neural code for natural scenes: from computations to neurons. Stanford Digital Repository. Available at https://purl.stanford.edu/rk663dm5577
```
thomasZen, line 73: Isn't that  duplicate information? Either just write:
```suggestion
  - Dataset available at https://purl.stanford.edu/rk663dm5577
```
Or at least make the style consistent with hoefling_2024. In both situations the paper describing the model and the paper describing the dataset is the same. And here we restate the paper, but in hoefling_2024 we only mention the link to the dataset.

thomasZen (approved):
Looks good. One suggestions, if you like the current version better go ahead with merging, though.

README.md
lhoefling: we never got a DOI for the hoefling 2024 dataset, hence we cannot "cite" it properly here. whereas the maheswaranathan dataset has a "preferred citation" section in the data repository (https://purl.stanford.edu/rk663dm5577). so yes it's duplicate but it seems that's just what it is like

PR #142, opened 2025-08-07
Feature: STSeparableBatchConv3d works with temporal_kernel_size < 3

thomasZen:


openretina/modules/layers/convolutions.py
```diff
@@ -383,9 +383,12 @@ def temporal_weights(
         Returns:
             tuple: Tuple containing sin and cos weights.
         """
-        K = length // 3
-        sin_weights = torch.nn.Parameter(data=torch.randn(num_feat, num_channels, K) * scale, requires_grad=True)
-        cos_weights = torch.nn.Parameter(data=torch.randn(num_feat, num_channels, K) * scale, requires_grad=True)
+        # From hoefling, 2024
+        # Then, in order to stay well under the Nyquist limit,
+        # we parameterize the kernels with k = 21/3 = 7 sines and cosines (that's why //3)
```
fededagos, line 418: I see now. But this assumes that we have 21 as the temporal kernel size, as it is now in the default config for Lara's model. Should we change this parametrisation somehow such that it does not rely on that implicitly?

thomasZen: I made it configurable.
Should we just set the subsampling value to 1 and also use this default for Lara's model? Or do you think it's more reasonable to keep it at 3 but do a fallback if `length < subsampling_factor`.

fededagos (approved):
This looks good. I believe you only need to sync again with main to resolve the torch version typing problems and then it should pass the tests.

PR #143, opened 2025-08-07
Documentation website

fededagos:
Use `mkdocs serve` locally to preview changes.
(have mkdocs installed first by doing `uv sync --extra dev`)


thomasZen (approved):
I think it would be good to merge this soon to avoid more merge conflicts in the future with regards to docstrings.

fededagos:
> I think it would be good to merge this soon to avoid more merge conflicts in the future with regards to docstrings.

As this is already public, I think we could merge it for now if you agree and then make fixes in follow-up commits. Would be nice to bump the minor version at least though if we do this.

PR #145, opened 2025-08-07
Sampled gaussian

Mvystrcilova:
Adds a FullGaussian2d readout as copied from neural predictors, along with its corresponding wrapper.
Also creates a new CoreReadout which includes a 3d ST core and this readout.

Extends the CoreReadout to allow for convolutions that are not sin,cos parametrized by setting a "type".
I am not sure this is the best way to do this, there could be another distinct CoreReadout class made for this but I felt like it would be too much overlap.

Also adds a single cell LNP.

### DIFF
--- a/openretina/models/core_readout.py
+++ b/openretina/models/core_readout.py
@@ -147,7 +147,7 @@ class BaseCoreReadout(LightningModule):
         # device = next(core.parameters()).device
 
         with torch.no_grad():
-            core_test_output = core.forward((torch.zeros((1,) + tuple(core_in_shape)), None))
+            core_test_output = core.forward(torch.zeros((1,) + tuple(core_in_shape)))
 
         return core_test_output.shape[1:]  # type: ignore
 
--- a/openretina/models/core_readout.py
+++ b/openretina/models/core_readout.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from typing import Any, Iterable, Optional, Literal
+from typing import Any, Iterable, Literal, Optional
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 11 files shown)


openretina/modules/core/base_core.py
```diff
@@ -116,23 +120,38 @@ def __init__(
...
-            )
-            layer["norm"] = torch.nn.BatchNorm3d(num_out_channels, momentum=0.1, affine=True)
-            layer["bias"] = Bias3DLayer(num_out_channels)
-            layer["nonlin"] = torch.nn.ELU()
-            if dropout_rate > 0.0:
-                layer["dropout"] = torch.nn.Dropout3d(p=dropout_rate)
-            if maxpool_every_n_layers is not None and (layer_id % maxpool_every_n_layers) == 0:
-                layer["pool"] = torch.nn.MaxPool3d((1, 2, 2))
-            self.features.add_module(f"layer{layer_id}", torch.nn.Sequential(layer))  # type: ignore
+            if type == 'sin_cos':
```
fededagos: We have something similar possible with the GRU core. We had a chat with Baptiste yesterday also about bringing the same functionality to the BaseCore, so thanks for this! I think though we might need to decide on a way to unify the naming between the two. @thomasZen thoughts on this?

openretina/modules/readout/gaussian.py
```diff
@@ -0,0 +1,408 @@
...
+from torch import nn
+import torch
+from torch.nn import functional as F
+
+from openretina.modules.readout.base import Readout
+
+
+class FullGaussian2d(Readout):
+    """
+    A readout using a spatial transformer layer whose positions are sampled from one Gaussian per neuron. Mean
```
fededagos, line 16: Just curious here: do you know where "spatial transformer" comes from? I have seen it a few times in neuralpredictors docs. Does it have anything to do with attention transformers?

openretina/modules/nonlinearities.py
fededagos, line 1: Is this also borrowed from neuralpredictors?

Mvystrcilova: The origial version that was in your code looks copied from there. All the regularizers acutally do. I changed it to not be just a convolution but a Laplace1d norm. Actually, it might make sense to leave the Laplace1d and then make a separate Laplace1dL2norm class.

openretina/modules/core/base_core.py
thomasZen: My thoughts are that we could merge the PR with the 'sin_cos' name for now, but create an issue to unify the names we use after the hackathon.
Regarding the name, maybe we either use the name of the class, or since STSeparableBatchConv3d basically learns the frequency components in the temporal component we reflect that in the name (maybe space_time_frequency_conv or sth. related?)

openretina/modules/core/base_core.py
```diff
@@ -68,6 +69,7 @@ def __init__(
         downsample_input_kernel_size: tuple[int, int, int] | None = None,
         input_padding: bool | int | str | tuple[int, int, int] = False,
         hidden_padding: bool | int | str | tuple[int, int, int] = True,
+        type='sin_cos'
```
thomasZen: I think here it's worth it to be verbose and all it convolution type
```suggestion
        convolution_type: str = 'sin_cos'
```

openretina/models/core_readout.py
```diff
@@ -144,10 +144,10 @@ def compute_readout_input_shape(
         self, core_in_shape: tuple[int, int, int, int], core: Core
     ) -> tuple[int, int, int, int]:
         # Use the same device as the core's parameters to avoid potential errors at init.
-        device = next(core.parameters()).device
+        # device = next(core.parameters()).device
 
         with torch.no_grad():
-            core_test_output = core.forward(torch.zeros((1,) + tuple(core_in_shape), device=device))
+            core_test_output = core.forward((torch.zeros((1,) + tuple(core_in_shape)), None))
```
thomasZen: "karamanlis_2024_gru" currently fails and I think it's because of this.
You can change this back to the original version to make it work again. If there's a reason to change this we can check for other solutions, though.

thomasZen (commented):
@Mvystrcilova I pushed a small commit to let the loading of old models still work: https://github.com/open-retina/open-retina/pull/145/commits/ff91b92a036a8fdd709dbbce29205a390aeb77d7 (the trick there is to make the buffer non persistent such that it does not get added to the state dict).
Model working is still not completely working because of changing the internals of the compute_readout_input_shape function, see the comment (i hope that's easy to fix though)

openretina/models/linear_nonlinear.py
```diff
@@ -93,3 +98,188 @@ def forward(self, x: Float[torch.Tensor, "batch channels t h w"], data_key: str
...
+            self.normalize_kernels()
+
+        out = self.space_conv(x)
+        out = self.time_conv(out)
+        out = self.nonlinearity(out)
+        out = rearrange(out, "batch neurons t 1 1 -> batch t neurons")
+        return out
+
+    def regularizer(self):
+        return self.laplace() + self.sparse_weight * self.weights_l1()
```
thomasZen:
```suggestion
        return self.laplace() + self.sparse_weight * self.weights_l1()

```
normally you end each file in a new line, but I think running `make fix-all` should apply the codestyle automatically and also add this new line.

openretina/modules/readout/multi_readout.py
```diff
@@ -94,6 +95,112 @@ def save_weight_visualizations(self, folder_path: str) -> None:
     def sessions(self) -> list[str]:
         return self.readout_keys()
 
+class MultiSampledGaussianReadoutWrapper(nn.ModuleDict):
+    """
+    Multiple Sessions version of the SimpleSpatialXFeature3d factorised gaussian readout.
```
thomasZen:
```suggestion
    Multiple Sessions version of the sampled gaussian readout.
```

thomasZen (approved):
From my side this looks okay once you fixed the errors in the test. I added comments that should point you in the right direction (I can help in case anything is not straight forward). I'm unsure if which code parts are based on neuralpredictors, etc. Putting docstrings that include pointer to neuralpredictors or also your paper in which you use the models is probably also a good idea :)

Thanks for the contribution!

PR #146, opened 2025-08-08
Greyscale model

SamuelSuhai:
I added some code to train a greyscale version of the Hoefling et al. 2024 model with the same data the chromatic version of the model was trained on. I added a `color_squashing_weights` argument of `SimpleCoreWrapper`. If this is None then the model will act unchanged. Otherwise the argument could be a tupple of weights, one per chromatic channel. The weights are used in the `WeightedChannelSumLayer` class. This class is used as the first layer to the core if the `color_squashing_weights` argument is not None. It then preforms a weighted sum across color channels, squashing this dimension to 1. It is therefore important that the the number of input-channels to the core is specified as 1 i.e. that the `channels` argument of `SimpleCoreWrapper` has value 1 on index 0 even if the training data has multiple color channels. This may be a bit counter intuitive. 

I used this layer to train a grey scale version of the hoefling_2024_core_readout_low_res model. With channel weights of 0.5 (basically applying an average to the two channel input) I got the following results: 
Correlation values for train, validation, and test dataloaders: 0.169, 0.234, 0.471. For the non-grey scale model these were: 0.180,0.241,0.495. 

Possible applications are: 
- using the hoefling et al. 2024 model for applications in which color is irrelevant
- comparing performance of a greyscale model vs a chromatic model across cell types.
- training models across datasets, some with greyscale some chromatic.

### DIFF
--- a/notebooks/training_hoefling_2024.ipynb
+++ b/notebooks/training_hoefling_2024.ipynb
@@ -725,21 +725,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "open_retina",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
-  },
-  "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.13.2"
   }
  },
  "nbformat": 4,
--- a/notebooks/training_hoefling_2024.ipynb
+++ b/notebooks/training_hoefling_2024.ipynb
@@ -725,9 +725,21 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "open_retina",
    "language": "python",
    "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.13.2"
   }
  },
  "nbformat": 4,
--- a/openretina/models/core_readout.py
+++ b/openretina/models/core_readout.py
@@ -200,7 +200,7 @@ class CoreReadout(BaseCoreReadout):
         dropout_rate: float = 0.0,
         maxpool_every_n_layers: Optional[int] = None,
         downsample_input_kernel_size: Optional[tuple[int, int, int]] = None,
-        color_squashing_weights: tuple[float] | None = None,
+        color_squashing_weights: tuple[float, ...] | None = None,
         data_info: dict[str, Any] | None = None,
     ):
         core = SimpleCoreWrapper(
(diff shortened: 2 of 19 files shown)


openretina/models/core_readout.py
```diff
@@ -200,6 +200,7 @@ def __init__(
         dropout_rate: float = 0.0,
         maxpool_every_n_layers: Optional[int] = None,
         downsample_input_kernel_size: Optional[tuple[int, int, int]] = None,
+        color_squashing_weights: tuple[float] | None = None,
```
thomasZen: I think this should be `tuple[float, ...]`
```suggestion
        color_squashing_weights: tuple[float, ...] | None = None,
```

notebooks/training_hoefling_2024.ipynb
```diff
@@ -725,21 +725,9 @@
...
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.13.2"
   }
  },
  "nbformat": 4,
```
thomasZen: Do not commit the updates to notebooks if they were not intentional or connected to the PR

openretina/modules/layers/reducers.py
```diff
@@ -0,0 +1,23 @@
+import torch
+import torch.nn as nn
+
+
+class WeightedChannelSumLayer(nn.Module):
+    """A layer that can reduced color inputs to greyscale by taking a weighted sum of the channels.
+    If the input is already greyscale, it will return the input unchanged."""
```
thomasZen: I'd first be abstract and only mention that it squashes the dimensions of the first channel, and then later provide the example of color-channels to greyscale. E.g. something like:
```suggestion
    """A layer that reduces the first channel of its input into one dimension according to the provided weights.
    If the first channel is already one dimensional, will return it unchanged.
    One use case is of this layer is to convert a multi-color channel into a grey-scale channel."""
```

openretina/modules/layers/reducers.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def __init__(self, init_channel_weights: tuple[float, ...], trainable: bool = False):
+        super().__init__()
+
+        # add the channel weights
+        self.channel_weights = nn.Parameter(torch.tensor(init_channel_weights), requires_grad=trainable)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """If the input is not greyscale, take a weighted sum to turn it greyscale."""
```
thomasZen:
```suggestion
        """If the input is not multi-dimensional, take a weighted sum to turn it greyscale."""
```

tests/modules/layers/test_reducers.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+import torch
+
+from openretina.modules.layers.reducers import WeightedChannelSumLayer
+
+
+@pytest.mark.parametrize(
+    "init_channel_weights, trainable",
+    [((5, 7), False), ((0.5, 0.5), False)],
```
thomasZen: Either add a test case for trainable=True or remove the argument.
```suggestion
    [((5, 7), False), ((0.5, 0.5), False), ((0.5, 0.5), True)],
```

thomasZen (commented):
Basically looks good, added smaller comments and get rid of the changes to the notebooks (I can help if you have problems there).
Nice PR! Ah, and it might make sense to post the results you got in the PR description.

thomasZen (approved):
Looks good.

Issue #148, opened 2025-08-14
compute_readout_input_shape runs on cpu


thomasZen:
compute_readout_input_shape that is called during initialization of the core_readout module seems to run on cpu, even if we load the model on cuda with  `load_core_readout_from_remote("hoefling_2024_base_low_res", "cuda")`.
Maybe this is because the model parameter only get mapped to cuda after loading it?
For context, the parameters are needed to figure out the shape of the output for the readout layers (as they for example initialize a grid of parameters dependent on the input height and width).

It probably makes sense to investigate this as for bigger model this runs a forward path and is compute heavy.
Maybe it even works to get rid of calling compute_readout_input_shape altogether by initializing the readout layers lazily, e.g. see:
- https://docs.pytorch.org/docs/stable/generated/torch.nn.LazyConv2d.html
- https://docs.pytorch.org/docs/stable/generated/torch.nn.modules.lazy.LazyModuleMixin.html#torch.nn.modules.lazy.LazyModuleMixin

thomasZen:
For the new UnfiedCoreReadout class (https://github.com/open-retina/open-retina/pull/154) this is resolved as it now stores the in_shape and only recalculates it when it is not passed:
https://github.com/open-retina/open-retina/blob/b3afc64116b775a0db45605eaab2d4f4e65bc55b/openretina/models/core_readout.py#L215-L222

PR #149, opened 2025-08-22
Bugfix: actually use torch conv3d, additonally pass convolution_type from config

thomasZen:
- Fixed bug that lead to an empty core for the convolution_type == "torch" (this was introduced in https://github.com/open-retina/open-retina/pull/145)
- Unified convolution type names by using get_conv_type for GRU and conv models
- Few consistency changes
- Exposed convolution_type in the config.

### DIFF
--- a/openretina/models/core_readout.py
+++ b/openretina/models/core_readout.py
@@ -358,6 +358,7 @@ class CoreGaussianReadout(BaseCoreReadout):
         dropout_rate: float = 0.0,
         maxpool_every_n_layers: Optional[int] = None,
         downsample_input_kernel_size: Optional[tuple[int, int, int]] = None,
+        convolution_type: str = "full",
         data_info: dict[str, Any] | None = None,
     ):
         core = SimpleCoreWrapper(
@@ -374,7 +375,7 @@ class CoreGaussianReadout(BaseCoreReadout):
             downsample_input_kernel_size=downsample_input_kernel_size,
             input_padding=core_input_padding,
             hidden_padding=core_hidden_padding,
-            convolution_type="torch",
+            convolution_type=convolution_type,
         )
 
         in_shape_readout = self.compute_readout_input_shape(in_shape, core)
--- a/openretina/models/core_readout.py
+++ b/openretina/models/core_readout.py
@@ -147,7 +147,8 @@ class BaseCoreReadout(LightningModule):
         device = next(core.parameters()).device
 
         with torch.no_grad():
-            core_test_output = core.forward(torch.zeros((1,) + tuple(core_in_shape)), device=device)
+            stimulus = torch.zeros((1,) + tuple(core_in_shape), device=device)
+            core_test_output = core.forward(stimulus)
 
         return core_test_output.shape[1:]  # type: ignore
 
--- a/openretina/models/core_readout.py
+++ b/openretina/models/core_readout.py
@@ -144,10 +144,10 @@ class BaseCoreReadout(LightningModule):
         self, core_in_shape: tuple[int, int, int, int], core: Core
     ) -> tuple[int, int, int, int]:
         # Use the same device as the core's parameters to avoid potential errors at init.
-        # device = next(core.parameters()).device
+        device = next(core.parameters()).device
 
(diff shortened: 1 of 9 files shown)


openretina/models/core_readout.py
```diff
@@ -144,10 +144,11 @@ def compute_readout_input_shape(
         self, core_in_shape: tuple[int, int, int, int], core: Core
     ) -> tuple[int, int, int, int]:
         # Use the same device as the core's parameters to avoid potential errors at init.
-        # device = next(core.parameters()).device
+        device = next(core.parameters()).device
```
thomasZen, line 147: this was commented out in a previous commit, I think with no good reason.
But see https://github.com/open-retina/open-retina/issues/148, this runs mostly on cpu anyway as it runs before we move the weights to gpu I think.

openretina/modules/core/base_core.py
```diff
@@ -124,48 +124,33 @@ def __init__(
...
-            elif self.convolution_type == "torch":
-                layer["conv"] = TorchSTSeparableConv3D(
-                    num_in_channels,
-                    num_out_channels,
-                    log_speed_dict={},
-                    temporal_kernel_size=temporal_kernel_sizes[layer_id],
-                    spatial_kernel_size=spatial_kernel_sizes[layer_id],
-                    bias=False,
-                    padding=padding,
-                )
```
thomasZen, line 164: this was a bug. For the elif, self.features.add_module was never called!
This lead to a readout only model with an empty core.

openretina/modules/layers/convolutions.py
```diff
@@ -440,3 +440,16 @@ def temporal_smoothing(sin: torch.Tensor, cos: torch.Tensor) -> torch.Tensor:
     reg = torch.sum((smoother * sin) ** 2) / F
     reg += torch.sum((smoother * cos) ** 2) / F
     return reg
+
+
+def get_conv_class(conv_type: str) -> type[nn.Module]:
```
thomasZen, line 453: this function was only moved

openretina/modules/nonlinearities.py
```diff
@@ -13,7 +13,7 @@ def forward(self, x):
         return torch.where(x > 0, self.c * x, self.a * (torch.exp(x / self.b) - 1))
 
 
-class parametrized_softplus(nn.Module):
+class ParametrizedSoftplus(nn.Module):
```
thomasZen, line 16: python style is to use camelCase

openretina/modules/core/base_core.py
```diff
@@ -189,11 +174,13 @@ def temporal_laplace(self) -> torch.Tensor:
...
     def temporal_smoothness(self) -> torch.Tensor:
-        if self.convolution_type == "torch":
+        if self.convolution_type == "separable":
             return self.temporal_laplace()
-        else:
+        elif self.convolution_type == "custom_separable":
             results = [temporal_smoothing(x.conv.sin_weights, x.conv.cos_weights) for x in self.features]
             return torch.sum(torch.stack(results))
+        else:
+            raise ValueError(f"Temporal smoothness not supported for {self.convolution_type=}")
```
thomasZen: I think we can implement this better in the future. Right now these losses only calculated if the loss weight > 0.0.
So it works if they the interpolation weights are set to 0.0 in the config.

We can rethink this in the future, though.

fededagos: Sounds good to me. Let's maybe explicitly put this in the error then.

```suggestion
            raise ValueError(
            f"Temporal smoothness not supported for {self.convolution_type=}. "
            "Set the temporal smoothness regularisation weight to 0 to still use this conv type."
            )
```

fededagos (approved):
Looks good to me! 
I would maybe only make the smoothness error more explicit before merging. 
Am I correct to think that this now unifies the conv_type convention between the normal core and the gru core?

thomasZen:
> Am I correct to think that this now unifies the conv_type convention between the normal core and the gru core?

Yes, that's true. I noticed that it was different and used that opportunity to unify it :)

PR #150, opened 2025-09-01
Goldin 2022 data, Klindt Readout, and Weight Visualization Callback

BaptisteLorenzi:
A bit of of a heavy branch. The goal was to have Image Models train properly (including hparams search) as needed in my lab. I was helped by an intern on that branch, not sure how to credit him.

- Implemented a Readout that is just a spatial filter + Feature Weights (As described in Klindt el al. 2018), this was lacking. Normally should also work in MultiReadout forms.
- Added data_io and configs for the goldin 2022, data> <!> I'm waiting for the PR on hugging face to be accepted to change the default path to the data in the data_io.
- Added a callback to visualize the model weights/filters during training. It saves all filter at each epoch by default but can be configured, very useful for debugging. It only properly work with BaseCore and KildtReadout, but it should be a good basis.

Missing : Currently the load_core_readout_model function only handles GRUCoreReadout and CoreReadout models.


configs/goldin_2022_core_readout.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - _self_ # values in this config will overwrite the defaults
+
+exp_name: core_readout_goldin_2022
+seed: 42
+check_stimuli_responses_match: false
+
+paths:
+  cache_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY} # Remote files are downloaded to this location
+  # If data_dir is a local path, data will be read from there. If a remote link, the target will be downloaded to cache_dir.
+  data_dir: '/home/baptiste/Documents/LabPipelines/open-retina/notebooks/data/omarre_lab/goldin_2022/'
```
thomasZen: Reminder: switch to remote path (and sorry for not reviewing the huggingface PR earlier)

configs/goldin_2022_core_readout.yaml
```diff
@@ -0,0 +1,65 @@
...
+  precision: 16-mixed
+
+training_callbacks:
+  early_stopping:
+    patience: 10
+    monitor: "val_loss"
+    mode: "min"
+  model_checkpoint:
+    monitor: "val_loss"
+    mode: "min"
```
thomasZen:
```suggestion
    monitor: "val_loss"
    mode: "min"

```

thomasZen: It's a bit nicer to end a file with a new line (both in the github interface and for linux specific reasons)

configs/goldin_2022_core_readout_hparams_search.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+exp_name: core_readout_goldin_2022
+seed: 42
+check_stimuli_responses_match: false
+
+paths:
+  cache_dir: ${oc.env:OPENRETINA_CACHE_DIRECTORY} # Remote files are downloaded to this location
+  # If data_dir is a local path, data will be read from there. If a remote link, the target will be downloaded to cache_dir.
+  data_dir: '/home/baptiste/Documents/LabPipelines/open-retina/notebooks/data/omarre_lab/goldin_2022/'
+  log_dir: "." # Used as parent for output_dir. Will store train logs.
```
thomasZen, line 28: reminder: remote path (sorry for not reviewing the huggingface PR earlier)

configs/goldin_2022_core_readout_hparams_search.yaml
```diff
@@ -0,0 +1,90 @@
...
+  in_shape: [1, 1, 108, 108]
+  hidden_channels: [4]
+  temporal_kernel_sizes: [1]
+  spatial_kernel_sizes: [30]
+  core_gamma_temporal: 0.0
+  core_gamma_input: 0.4654576767872573
+  convolution_type: "time_independent"
+  mask_l1_reg : 5.924553274051563
+  weights_l1_reg : 6.075840974162483
+  laplace_mask_reg: 1.706070712749228
```
thomasZen, line 70: out of interest,  are these the best values found with hyperparameter search? Otherwise, why are they so specific?

configs/goldin_2022_core_readout_hparams_search.yaml
```diff
@@ -0,0 +1,90 @@
...
+  precision: 16-mixed
+
+training_callbacks:
+  early_stopping:
+    patience: 10
+    monitor: "val_loss"
+    mode: "min"
+  model_checkpoint:
+    monitor: "val_loss"
+    mode: "min"
```
thomasZen:
```suggestion
    mode: "min"

```
new line at end of file

configs/training_callbacks/weight_visualization.yaml
```diff
@@ -0,0 +1,9 @@
+# # Visualize frequently at the beginning, then less often
+weight_visualization:
+  _target_: openretina.callbacks.weight_visualization_callback.ConditionalWeightVisualizationCallback
+  every_n_epochs: 1
+  start_epoch: 0
+  call_on_train_epoch_end: true
+  call_on_validation_epoch_end: false
+  save_plots: true
+  folder_path: "${paths.output_dir}/weight_plots/"
```
thomasZen:
```suggestion
  folder_path: "${paths.output_dir}/weight_plots/"

```
new line at end of file

notebooks/hoefling_gradient_analysis.ipynb
```diff
@@ -273,10 +273,10 @@
...
-    "    x, x, \n",
-    "    gradient_grid, \n",
+    "    x, x,\n",
+    "    gradient_grid,\n",
     "    resp_grid,\n",
-    "    normalize_response=True, \n",
+    "    normalize_response=True,\n",
     "    rc_dict=rc_dict\n",
     ")"
    ]
```
thomasZen, line 282: These changes were done automatically by your IDE I suppose?
If it's very easy we could reset those as this PR didn't touch the notebooks (I'm sympathetic with going over all notebooks in a separate PR to apply this consistently though if it's a IDE thing)

thomasZen: also applies to notebooks below that have similar "codestyle" changes

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __init__(
+        self,
+        every_n_epochs=1,
+        start_epoch=0,
+        call_on_train_epoch_end=True,
+        call_on_validation_epoch_end=False,
+        save_plots=False,
+        folder_path=None,
+    ):
```
thomasZen:
```suggestion
    def __init__(
        self,
        every_n_epochs: int = 1,
        start_epoch: int = 0,
        call_on_train_epoch_end: bool = True,
        call_on_validation_epoch_end: bool = False,
        save_plots: bool = False,
        folder_path: str | None = None,
    ):
```
If we put the type information from the docstring as type annotations these will be automatically checked for consistency.

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        super().__init__()  # Explicit parent call
+        self.every_n_epochs = every_n_epochs
+        self.start_epoch = start_epoch
+        self.call_on_train_epoch_end = call_on_train_epoch_end
+        self.call_on_validation_epoch_end = call_on_validation_epoch_end
+        self.save_plots = save_plots
+        self.folder_path = folder_path
+
+    def _should_visualize(self, current_epoch):
```
thomasZen:
```suggestion
    def _should_visualize(self, current_epoch: int) -> bool:
```

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+
+@contextmanager
+def eval_state(model: torch.nn.Module):
+    training_status = model.training
+
+    try:
+        model.eval()
+        yield model
+    finally:
+        model.train(training_status)
```
thomasZen: This is a duplicate of https://github.com/open-retina/open-retina/blob/d5786868e557ff3bdd21a006af855596b12ec65a/openretina/utils/model_utils.py#L7-L15, so just importing the function is better here.

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class ConditionalWeightVisualizationCallback(Callback):
+    """
+    PyTorch Lightning callback that calls the model's save_weight_visualizations
+    function conditionally based on epoch settings.
+    """
```
thomasZen: Awesome!

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+        self.call_on_train_epoch_end = call_on_train_epoch_end
+        self.call_on_validation_epoch_end = call_on_validation_epoch_end
+        self.save_plots = save_plots
+        self.folder_path = folder_path
+
+    def _should_visualize(self, current_epoch):
+        """Check if we should visualize at the current epoch."""
+        return current_epoch >= self.start_epoch and current_epoch % self.every_n_epochs == 0
+
+    def _call_visualization(self, trainer, pl_module, stage=""):
```
thomasZen:
```suggestion
    def _call_visualization(self, trainer, pl_module, stage: str = "") -> None:
```

thomasZen (commented):
Will finish today / tomorrow morning

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+                    )
+            pl_module.save_weight_visualizations(**kwargs)
+
+        except Exception as e:
+            print(f"Warning: Failed to plot weight visualization at epoch {trainer.current_epoch}: {e}")
+
+    def on_train_start(self, trainer, pl_module):
+        if trainer.current_epoch == 0:
+            # print("Running weight visualization on train start")
+            self._call_visualization(trainer, pl_module, "init")
```
thomasZen:
```suggestion
            self._call_visualization(trainer, pl_module, "init")
```
remove print statement if it's not used

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def on_train_epoch_end(self, trainer, pl_module):
+        if self.call_on_train_epoch_end and self._should_visualize(trainer.current_epoch):
+            self._call_visualization(trainer, pl_module, "train")
+
+    def on_validation_epoch_end(self, trainer, pl_module):
+        if self.call_on_validation_epoch_end and self._should_visualize(trainer.current_epoch):
+            self._call_visualization(trainer, pl_module, "val")
```
thomasZen, line 74: I think this is fine, but what's the motivation of having both of these (ie on_train_epoch_end and on_validation_epoch_end)? It makes some workflows a bit messy, e.g. theoretically there could be setups for which you run validation multiple times per epoch if your training corpus is really big.

configs/training_callbacks/weight_visualization.yaml
```diff
@@ -0,0 +1,9 @@
+# # Visualize frequently at the beginning, then less often
```
thomasZen: That's not what's happening, right? This config specifies that we validate every epoch if I'm not mistaken.

openretina/callbacks/weight_visualization_callback.py
thomasZen: Or convert to a comment without if you want to keep that information:
```suggestion
            # Running weight visualization on train start
            self._call_visualization(trainer, pl_module, "init")
```

openretina/callbacks/weight_visualization_callback.py
```diff
@@ -0,0 +1,96 @@
...
+                    kwargs["state_suffix"] = f"{stage}_epoch_{trainer.current_epoch}"
+                else:
+                    print(
+                        "Weights visualizations will only save the last epoch. "
+                        "For epoch by epoch visualization, implement state_suffix in the viz function"
+                    )
+            pl_module.save_weight_visualizations(**kwargs)
+
+        except Exception as e:
+            print(f"Warning: Failed to plot weight visualization at epoch {trainer.current_epoch}: {e}")
```
thomasZen: I would restructure this, as I think:
- Currently, it's confusing behavior if self.save_plots = True and self.folder_path = None (or self.folder_path = "") as then the weights do not get visualized.
- If the `if self.save_plots and self.folder_path:` path is not taken, the `pl_module.save_weight_visualizations` will be called with kwargs={}, which will raises an exception. I think this is not intended?

thomasZen: One suggestion to fix that is to get rid of the `self.save_plots` argument completely, and return early if self.folder_path is None, e.g.:
```suggestion
        if not self.folder_path:
            return
        if not hasattr(pl_module, "save_weight_visualizations"):
            print("Warning: Model does not have 'save_weight_visualizations' method")
            return

        kwargs = {}

        # Handle folder_path if requested
        try:
            os.makedirs(self.folder_path, exist_ok=True)

            # Check if the function accepts folder_path parameter
            sig = inspect.signature(pl_module.save_weight_visualizations)
            if "folder_path" in sig.parameters:
                kwargs["folder_path"] = self.folder_path
            if "state_suffix" in sig.parameters:
                kwargs["state_suffix"] = f"{stage}_epoch_{trainer.current_epoch}"
            else:
                print(
                    "Weights visualizations will only save the last epoch. "
                    "For epoch by epoch visualization, implement state_suffix in the viz function"
                )
            pl_module.save_weight_visualizations(**kwargs)

        except Exception as e:
            print(f"Warning: Failed to plot weight visualization at epoch {trainer.current_epoch}: {e}")
```

But feel free to use a different implementation (I hope I understood the purpose of this function correctly)

openretina/data_io/goldin_2022/responses.py
```diff
@@ -0,0 +1,62 @@
+"""
+Minimal stimulus loading utilities to train a model on the data used in Maheswaranathan et al. 2023
```
thomasZen: this has to get updated, too.

openretina/data_io/goldin_2022/stimuli.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+
+import os
+from typing import Literal
+
+from openretina.data_io.base import MoviesTrainTestSplit, normalize_train_test_movies
+from openretina.utils.file_utils import get_local_file_path
+from openretina.utils.h5_handling import load_dataset_from_h5
+
+CLIP_LENGTH = 1  # one frame of 300ms
```
thomasZen: this was already define in goldin_2022/constants.py (so import this here or delete the constants module)

openretina/data_io/goldin_2022/stimuli.py
```diff
@@ -0,0 +1,70 @@
...
+        # Add channel dimension
+        train_image = train_image[None, ...]
+        test_image = test_image[None, ...]
+
+        if normalize_stimuli:
+            train_image, test_image, norm_dict = normalize_train_test_movies(train_image, test_image)
+        else:
+            norm_dict = {"norm_mean": None, "norm_std": None}
+
+        stimuli_all_sessions[str(session).replace(".h5", "")] = MoviesTrainTestSplit(
```
thomasZen: There's [removesuffix](https://docs.python.org/3/library/stdtypes.html#str.removesuffix) which is closer to what you want to do here I think:
```suggestion
        stimuli_all_sessions[str(session).removesuffix(".h5")] = MoviesTrainTestSplit(
```

openretina/data_io/goldin_2022/stimuli.py
```diff
@@ -0,0 +1,70 @@
...
+            train=train_image,
+            test=test_image,
+            stim_id=stim_type,
+            random_sequences=None,
+            norm_mean=norm_dict["norm_mean"],
+            norm_std=norm_dict["norm_std"],
+        )
+    print(stimuli_all_sessions.keys())
```
thomasZen: Should we keep this print statement?

openretina/models/core_readout.py
```diff
@@ -403,6 +428,76 @@ def __init__(
         self.save_hyperparameters()
 
 
+class CoreKlindtReadout(BaseCoreReadout):
```
thomasZen, line 431: After you merged this PR, do you have any objections that I refactor these classes to add readout_type: str as an optional parameter, and merge some classes (it will lead to more code duplication if we have a new class for every new readout type)

openretina/modules/core/base_core.py
```diff
@@ -224,12 +227,14 @@ def plot_weight_visualization(self, layer: int, in_channel: int, out_channel: in
...
+    def save_weight_visualizations(
+        self, folder_path: str, file_format: str = "jpg", state_suffix: Optional[str] = None
+    ) -> None:
         for i, layer in enumerate(self.features):
             output_dir = os.path.join(folder_path, f"weights_layer_{i}")
             os.makedirs(output_dir, exist_ok=True)
-            layer.conv.save_weight_visualizations(output_dir, file_format)
-            print(f"Saved weight visualization at path {output_dir}")
+            layer.conv.save_weight_visualizations(output_dir, file_format, state_suffix)
+            # print(f"Saved weight visualization at path {output_dir}")
```
thomasZen: Either remove completely or actually print.

openretina/modules/layers/convolutions.py
```diff
@@ -144,6 +146,10 @@ def __init__(
             padding=padding,
             bias=bias,
         )
+        self.weight_spatial = self.conv.weight.data
+
+    def refresh_spatial_weight_attribute(self):
+        self.weight_spatial = self.conv.weight.data
```
thomasZen:
```suggestion

    @property
    def weight_spatial(self):
        self.weight_spatial = self.conv.weight.data
```
I think using @property has the same effect and we have less state in this class (e.g. we do not have to think about updating self.weight_spatial at all), see: https://docs.python.org/3/library/functions.html#property

openretina/modules/readout/klindt_readout.py
```diff
@@ -0,0 +1,200 @@
...
+        weights_constraint: Optional[str] = None,
+        mask_constraint: Optional[str] = None,
+        init_mask: Optional[torch.Tensor] = None,
+        init_weights: Optional[torch.Tensor] = None,
+        init_scales: Optional[Sequence[Tuple[float, float]]] = None,
+    ):
+        """
+        TO DO:
+        - Refactorised some names
+        """
```
thomasZen: Spurious todo comment, delete?

thomasZen: a docstring here would also be useful, e.g. for the weights_constraint and mask_constraints options.

openretina/modules/readout/klindt_readout.py
```diff
@@ -0,0 +1,200 @@
...
+        self._input_weights_regularizer_spatial = FlatLaplaceL23dnorm(padding=0)
+        self.outdims = num_neurons
+
+        if isinstance(mask_size, int):
+            self.num_mask_pixels = mask_size**2
+            self.mask_size = (mask_size, mask_size)
+        else:
+            h, w = mask_size
+            self.num_mask_pixels = h * w
+            self.mask_size = mask_size
```
thomasZen: I think you only use num_masks pixels in the init function. If so keep it as a local variable. I.e. use `num_mask_pixels = h * w` instead of using `self.num_mask_pixels = h * w`).

(The reason is that every time you add state you add opportunities to make the state inconsistent. E.g. somebody could update `self.mask_size` while forgetting to update `self.num_mask_pixels`, so better to avoid this error type completely whenever possible)

thomasZen (commented):
More, but looks nice so far!

openretina/data_io/goldin_2022/responses.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # Load neural responses
+        train_session_data = load_dataset_from_h5(recording_file, f"/train/response/{response_type}")
+        test_session_data = load_dataset_from_h5(recording_file, f"/test/response/{response_type}")
+
+        assert train_session_data.shape[0] == test_session_data.shape[0], (
+            "Train and test responses should have the same number of neurons."
+        )
+
+        responses_all_sessions[str(session).replace(".h5", "")] = ResponsesTrainTestSplit(
```
thomasZen: maybe:
```suggestion
        responses_all_sessions[str(session).removesuffix(".h5")] = ResponsesTrainTestSplit(
```
see: https://docs.python.org/3/library/stdtypes.html#str.removesuffix

openretina/modules/readout/klindt_readout.py
```diff
@@ -0,0 +1,200 @@
...
+    def forward(self, x: torch.Tensor, **kwargs: Any) -> torch.Tensor:
+        self.apply_constraints()
+        B, C, T, H, W = x.shape
+        h, w = self.mask_size
+        assert H == h and W == w
+        x_flat = x.view(B, C, T, -1).permute(0, 2, 1, 3)
+        masked = torch.matmul(x_flat, self.mask_weights)
+        masked = masked.permute(0, 1, 3, 2)
+        out = (masked * self.readout_weights.T.unsqueeze(0).unsqueeze(0)).sum(dim=3)
+        return F.softplus(out + self.bias) if self.readout_bias else out
```
thomasZen: If there's no bias, we do not use an activation function?

thomasZen (commented):
Looks good so far, thanks for adding the PR. I'm especially excited about using the weight visualizations during training, and the new data set :)
Most of the comments are small, but there are some possibilities to make the code more robust for the future.

I'll try out the PR after you made a round of changes and then approve if it works!

thomasZen:
> I was helped by an intern on that branch, not sure how to credit him.

How about adding him in the citations file of this repository? (the plan is to update it before the ERM I think)

fededagos:
An additional option to credit other people having worked on this would be to let them make at least one commit on this PR, so that they will also show as contributors on the main github page of the project :)

configs/goldin_2022_core_readout_hparams_search.yaml
BaptisteLorenzi: They are good parameters I found with hparams search while testing this PR around. Anyway, if you use the --multirun mode, it will overwrite those with the values for the current trial.

openretina/callbacks/weight_visualization_callback.py
BaptisteLorenzi: The function is not used in this file anymore so I just removed it.

openretina/callbacks/weight_visualization_callback.py
BaptisteLorenzi: The validation one is set to False by default. I agree that we can just remove it if you want to keep the workflow as simple as possible.

configs/training_callbacks/weight_visualization.yaml
BaptisteLorenzi: Comment removed

openretina/callbacks/weight_visualization_callback.py
BaptisteLorenzi: Your implementation makes more sense providing we are expecting 'folder_path' to always be an argument of the module.save_weight_visualizations .If so, I can simplify and correct the function using your remarks.

openretina/models/core_readout.py
BaptisteLorenzi: No, on the contrary I think this would be very welcome. I just didn't know how to do it properly. The config files will have to be updated accordingly.

openretina/modules/readout/klindt_readout.py
BaptisteLorenzi: No, this is an error due to the fact we never used the no bias option. I changed it to be : 
```
if self.readout_bias:
            out = out + self.bias
return F.softplus(out)
```
        
In fact there is probably a better way to handle this, mainly by avoiding creating the bias parameter if readout_bias == False. But that should be minor.

BaptisteLorenzi:
> An additional option to credit other people having worked on this would be to let them make at least one commit on this PR, so that they will also show as contributors on the main github page of the project :)

He is no longer working with us. It was a short summer internship. I'll contact him to see if he can just do an empty commit. As for the citation file, I'm not sure what you mean.

BaptisteLorenzi:
Hello, I've handled most comments. Ther were really useful. I think a short second pass is needed.
On the topic of weight visualization, keep in mind that there is probably a little bit of work to do to make it work with the plot function from other modules. But this should be quite simple.
Finally, I'm considering pushing some trained versions of the 2D image model. But I don't really know what to put in the data_info part. It would be nice if we could conclude on a 'user guide' on how to save models and 'what' to save with them. (This would be useful to use this model for the gradient visualization method demo).

configs/goldin_2022_core_readout_hparams_search.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - lr_monitor
+    - model_checkpoint
+    - weight_visualization
+  - logger:
+    - tensorboard
+    - csv
+    - mlflow
+  - trainer: default_deterministic
+  - hydra: default
+  - override hydra/sweeper: optuna
```
thomasZen: When running this config I got the error message:
```
In 'hydra/config': Could not find 'hydra/sweeper/optuna'
```

After installing optuna with `pip install hydra-optuna-sweeper` it worked though.
We should probably:
- Add optuna as optional requirements.
- Adapt the config so that it either works by default or add a comment here that you have to install optuna

(also open to other options if you have ideas)

thomasZen: I did above

thomasZen (approved):
Looks good, approved.

I additionally added hydra-optuna-sweeper as an dependency, cleaned up the weight visualization class a bit, and updated some save_weight_visualization function with state_suffix (I changed it to always be a string, too).

I also tested that training on goldin data works, which is awesome :) The weight visualization callback now also works for the low res hoefling model (it's just slow as it has a lot of neurons).

Feel free to do a further pass if you want and/or merge yourself whenever you are ready.

thomasZen:
> Finally, I'm considering pushing some trained versions of the 2D image model. But I don't really know what to put in the data_info part. It would be nice if we could conclude on a 'user guide' on how to save models and 'what' to save with them.

Sounds good. I'll talk to Federico today about a user guide for the data_info part, and try to get this started.

PR #94, opened 2025-01-30
ROI view notebook demo

fededagos:
This PR adds a "demo" notebook with various visualizations as an interactive entrypoint to some of the package functionalities. 
Main functionalities are:

- Downloading stimuli
- Loading a pre-trained model
- Make predictions on test stimuli
- Visualize activations with an interactive ROI view.

This could be extended in other ways, for example showing any arbitrary video to the retina model, which might also be cool

PR #154, opened 2025-09-12
UnifiedCoreReadout class that instantiates the core and the readout separately from hydra

thomasZen:
Specify both core and readout classes to use in the config via hydra. This makes it possible to use any core-readout combination without any code changes.
To make this work, it was necessary to unify the arguments for the classes a bit, i.e.:
- The core now always get `channels` and `n_neurons_dict` as arguments
- The readout always gets `in_shape: tuple[int, int, int, int]` as an argument

I also scratched support for the current pre-trained GRU models on hugginface (I think that's reasonable as I don't know of anybody who uses them right now). The other models are still loadable, though.

I tested `openretina train --config-path configs --config-name ` with the following configs so far (i.e., checked if start training  without an error):
- [x] goldin_2022_core_readout_hparams_search
- [x] hdf5_core_readout
- [x] hoefling_2024_core_readout_high_res
- [x] hoefling_2024_core_readout_low_res_hyperparams_search
- [x] hoefling_2024_core_readout_low_res
- [x] karamanlis_2024_core_readout
- [x] maheswaranathan_2023_core_readout
- [x] maheswaranathan_2023_LNP (uses the old style though, could adapt that in a future PR)


openretina/cli/train.py
```diff
@@ -71,18 +71,20 @@ def train_model(cfg: DictConfig) -> float | None:
...
     else:
         # Assign missing n_neurons_dict to model
         cfg.model.n_neurons_dict = data_info["n_neurons_dict"]
-        log.info(f"Instantiating model <{cfg.model._target_}>")
-        model = hydra.utils.instantiate(cfg.model, data_info=data_info)
+        if hasattr(cfg.model, "_target_"):
+            log.info(f"Instantiating model <{cfg.model._target_}>")
+            model = hydra.utils.instantiate(cfg.model, data_info=data_info)
+        else:
+            log.info("Instantiating model <UnifiedCoreReadout>")
```
fededagos, line 86: It would be cool to add some additional logging saying what the core type and readout type are, but I guess it is not so necessary given that we usually always print the full config during logging as the first thing

openretina/models/core_readout.py
```diff
@@ -198,7 +195,39 @@ def update_model_data_info(self, data_info: dict[str, Any]) -> None:
...
+        # determine input_shape of readout if it is not already present
+        if "in_shape" not in readout:
+            in_shape_readout = self.compute_readout_input_shape(in_shape, core_module)
+            readout["in_shape"] = (in_shape_readout[0],) + in_shape_readout[1:]
+        readout_module = hydra.utils.instantiate(
+            readout,
+            n_neurons_dict=n_neurons_dict,
+        )
+
+        # if calling save_hyperparameters after __init__ it leads to errors related to data_info["session_kwargs"]
```
fededagos, line 224: Did you check whether all the entries are saved correctly though? i.e. whether data info holds everything correctly, and so on

fededagos (commented):
Looks good! I think the only important thing to check before merging is whether data info and the hyperparameters are saved correctly by calling save_hyperparameters before the __init__. 
Otherwise, this is very cool. Much more flexible and modular than what we had before!

fededagos (approved):

openretina/models/core_readout.py
thomasZen: I checked, it's basically the same module:
- input_shape is now a tuple instead of a list
- norm_dict has np.nan instead of None as values (this comes from https://github.com/open-retina/open-retina/pull/94/files#r2358106258)

Both is unrelated to this PR, though, and related to previous code changes.

openretina/cli/train.py
thomasZen: Yeah, for training we print this right before starting to train:
```
  | Name             | Type                        | Params | Mode 
-------------------------------------------------------------------------
0 | core             | SimpleCoreWrapper           | 12.4 K | train
1 | readout          | MultiGaussianReadoutWrapper | 66.0 K | train
2 | loss             | PoissonLoss3d               | 0      | train
3 | correlation_loss | CorrelationLoss3d           | 0      | train
-------------------------------------------------------------------------
```
So I think the log message above is good enough and the rest can be read out here.

PR #156, opened 2025-09-18
Add uv for project and package management

fededagos:


README.md
```diff
@@ -14,17 +14,20 @@ The models in this repository are inspired by and partially contain adapted code
...
 cd open-retina
-pip install -e .
+uv sync
 ```
 
 For normal usage:
 
 ```
-pip install openretina
+uv pip install openretina
```
thomasZen: I'm worried that we loose some people if we only talk about uv.
Maybe it makes sense to restructure this section a bit, e.g.:


Openretina supports installations via pip and [uv](https://docs.astral.sh/uv/).
If you plan to use openretina with pretrained models there's no need to check out the Github repository, you can just install the openretina package as follows:
```
# if you prefer using pip
pip install openretina
# if you prefer using uv
uv pip install openretina
```

If you want to train your own models, want to run Jupyter notebooks, or change the code of openretina, we recommend using `uv`, a fast and flexible project and package manager. 
If you are not familiar with uv, check out their [simple quickstart guide](https://docs.astral.sh/uv/).

```
git clone git@github.com:open-retina/open-retina.git
cd open-retina
pip install -e .
uv sync --extra dev
```

Maybe also adding --extra dev for the more development focused installation makes sense, too.

thomasZen (approved):
Looks good. I added a suggestion to update the README to both talk about pip and uv for users that only want to use openretina for pretrained models as I assume not all of them might be willing to learn about uv.

PR #158, opened 2025-09-22
New field standards for `data_info`

### LINKED ISSUE #153: New `data_info` standard
Following discussions at the workshop, we need to implement a new `data_info` standard. Here is a proposal based on previous discussions. @thomasZen and @BaptisteLorenzi: please chime in if you have further input, I would start implementing the necessary changes to save these from next week.

First, I would propose to make the information also available under the alias `pretrained_cfg`, i.e. following similar conventions as [torch image models (`timm`)](https://huggingface.co/docs/timm/en/quickstart).

Then, I would propose saving the following fields in the dictionary:
- `n_neurons` : Dictionary, with sessions names as keys, and number of neurons in the sessions as values. This is already present.
- `mean`: tuple, with mean normalisation values for the videos. Would replace "movie_norm_dict" and "norm_mean" inside that, for ease of access.
- `std`: tuple, same as above. The idea is to flatten a bit the entries and make them more accessible
- `input_shape`: already present. Shape of the video input, excluding time dimension
- `response_rate`: float. Sampling rate of responses (i.e. the one used to train, so does not need to equate original experimental sampling rate for example after upsampling or downsampling from binning).
- `stimulus_rate`: float . Frame rate of the stimulus.
- `stimulus_class`: str, or list of str. The type of stimuli used to train the model, e.g. white noise, natural movies,... I do not see a need to standardise this right now, but we could settle on literal entries to use here. 
- `model_cut_frames`: int. Stimulus frames cut by the model (e.g. as a consequence of convolutions).
- `conv_type`: optional. Entry would be a string, indicating the convolution type the model is using.
- `core_type`: optional. Entry would be a string, indicating the type of core the model is using. 
- `readout_type`: optional. Entry would be a string, indicating which readout class is being used.
- `sessions_info`: optional. Nested dictionary with all extra session information that can be useful for downstream analyses. Keys would be session names, and values other dictionaries containing information either globally for the session (e.g. like the eye), or arrays containing information per neuron. This would be a potential candidate for standardisation.

I would vote for keeping the explainable variance per neuron / model performance of a neuron optional, and potentially going into sessions_info. I think it would still be useful to have an utility function that automatically adds at least the model performance at the end of training. I can add this in the same or a separate PR.

fededagos:
Fixes #153 . See issue for details.

### DIFF
--- a/configs/data_io/hoefling_2024.yaml
+++ b/configs/data_io/hoefling_2024.yaml
@@ -37,7 +37,7 @@ responses:
         cache_folder: ${paths.cache_dir}
 
 data_info:
-  response_rate: 30
-  stimulus_rate: 30
+  response_rate_hz: 30
+  stimulus_rate_hz: 30
   stimulus_class: ${data_io.responses.response_type}
   retina_pixel_size_um: 12.5
--- a/configs/data_io/hoefling_2024.yaml
+++ b/configs/data_io/hoefling_2024.yaml
@@ -4,7 +4,7 @@ stimuli:
   file_path:
     _target_: openretina.utils.file_utils.get_local_file_path
     _convert_: object
-    file_path: ${paths.movies_path} # Reads from the main config. Can be both higgh and low res.
+    file_path: ${paths.movies_path} # Reads from the main config. Can be both high and low res.
     cache_folder: ${paths.cache_dir}
 
 # Taking hydra recursive instantiation to the extreme
@@ -35,3 +35,9 @@ responses:
         _convert_: object
         file_path: ${paths.responses_path}
         cache_folder: ${paths.cache_dir}
+
+data_info:
+  response_rate: 30
+  stimulus_rate: 30
+  stimulus_class: ${data_io.responses.response_type}
+  retina_pixel_size_um: 12.5
--- a/openretina/cli/train.py
+++ b/openretina/cli/train.py
@@ -74,11 +74,11 @@ def train_model(cfg: DictConfig) -> float | None:
         model = load_core_readout_model(load_model_path, device)
 
         # add new readouts and modify stored data in model
-        model.readout.add_sessions(data_info["n_neurons_dict"])  # type: ignore
+        model.readout.add_sessions(data_info["n_neurons"])  # type: ignore
         model.update_model_data_info(data_info)
     else:
         # Assign missing n_neurons_dict to model
-        cfg.model.n_neurons_dict = data_info["n_neurons_dict"]
+        cfg.model.n_neurons_dict = data_info["n_neurons"]
         if hasattr(cfg.model, "_target_"):
             log.info(f"Instantiating model <{cfg.model._target_}>")
             model = hydra.utils.instantiate(cfg.model, data_info=data_info)
--- a/openretina/cli/train.py
+++ b/openretina/cli/train.py
@@ -53,7 +53,7 @@ def train_model(cfg: DictConfig) -> float | None:
         movies_dictionary=movies_dict,
     )
 
-    data_info = compute_data_info(neuron_data_dict, movies_dict)
+    data_info = compute_data_info(neuron_data_dict, movies_dict, partial_data_info=cfg.data_io.get("data_info"))
 
     train_loader = data.DataLoader(
         LongCycler(dataloaders["train"], shuffle=True),
@@ -114,8 +114,8 @@ def train_model(cfg: DictConfig) -> float | None:
     dataloader_mapping = {f"DataLoader {i}": x[0] for i, x in enumerate(short_cyclers)}
     log.info(f"Dataloader mapping: {dataloader_mapping}")
     trainer.test(model, dataloaders=[c for _, c in short_cyclers], ckpt_path="best")
-    # Check if MLflow is one of the loggers and save model and datasets as artifacts
 
+    # Check if MLflow is one of the loggers and save model and datasets as artifacts
     mlflow_logger_array = [logger for logger in logger_array if "mlflow" in str(type(logger)).lower()]
     if len(mlflow_logger_array) > 1:
         raise ValueError(f"Multiple mlflow loggers defined:  {[str(type(logger)) for logger in mlflow_logger_array]}")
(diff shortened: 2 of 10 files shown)


fededagos:
Right now this is working and tested only on hoefling_2024 models.

Outstanding TODOs: 

- [ ] Add pixel size and other metadata for datasets other than hoefling_2024
- [ ] Retrain and upload all models with updated data_info to huggingface

openretina/cli/train.py
```diff
@@ -53,7 +53,7 @@ def train_model(cfg: DictConfig) -> float | None:
         movies_dictionary=movies_dict,
     )
 
-    data_info = compute_data_info(neuron_data_dict, movies_dict)
+    data_info = compute_data_info(neuron_data_dict, movies_dict, partial_data_info=cfg.data_io.get("data_info"))
```
thomasZen, line 56: Does it work if partial_data_info is None? (`.get("data_info")` returns None if "data_info" is not in the dict)

openretina/data_io/base.py
```diff
@@ -179,33 +179,58 @@ def normalize_train_test_movies(
...
+
+        input_shape = input_shapes[0]
+
     sessions_kwargs = {
         session_name: responses.session_kwargs for session_name, responses in neuron_data_dictionary.items()
     }
 
     return {
-        "n_neurons_dict": n_neurons_dict,
+        "n_neurons": n_neurons_dict,
```
thomasZen: any reason for this rename?

openretina/models/core_readout.py
```diff
@@ -194,6 +220,11 @@ def update_model_data_info(self, data_info: dict[str, Any]) -> None:
             self.hparams["n_neurons_dict"] = self.data_info["n_neurons_dict"]
             self.hparams["data_info"] = self.data_info
 
+    @property
+    def pretrained_cfg(self) -> dict[str, Any]:
+        """Alias for data_info, following timm conventions."""
```
thomasZen: maybe add a longer name or a link to the timm convention.
The first link that comes up on google is this one: https://www.timm2025.org/

openretina/models/core_readout.py
```diff
@@ -194,6 +220,11 @@ def update_model_data_info(self, data_info: dict[str, Any]) -> None:
             self.hparams["n_neurons_dict"] = self.data_info["n_neurons_dict"]
```
thomasZen, line 220: is this still up-to-date? There was a diff renaming "n_neurons_dict" to "n_neurons":
```
-        cfg.model.n_neurons_dict = data_info["n_neurons_dict"]
+        cfg.model.n_neurons_dict = data_info["n_neurons"]
```

openretina/modules/losses/correlation.py
```diff
@@ -41,6 +51,11 @@ def forward(self, output, target, group_assignment, group_counts):
             (var_out + self.eps) * (var_target + self.eps)
         ).sqrt()
         corrs = corrs * (group_counts.sum() / group_counts.size(0) / group_counts[group_assignment - 1])[:, None, ...]
+
+        per_neuron_correlations = -corrs.view(-1, corrs.shape[-1]).mean(dim=0)
+        self._per_neuron_correlations = per_neuron_correlations.detach()
```
thomasZen, line 56: maybe add `.cpu().numpy()`, too. (I slightly prefer numpy over torch arrays for metadata like things)

thomasZen: but if doing though you'd have to adapt code below, too. Maybe just using `cpu()` or leaving as is is also good.

configs/data_io/hoefling_2024.yaml
```diff
@@ -35,3 +35,9 @@ responses:
         _convert_: object
         file_path: ${paths.responses_path}
         cache_folder: ${paths.cache_dir}
+
+data_info:
+  response_rate: 30
+  stimulus_rate: 30
```
thomasZen: if it's easy to change:
```suggestion
  response_rate_hz: 30
  stimulus_rate_hz: 30
```
as we mention units for the pixel size, too.

thomasZen (approved):
Looks good, added a few suggestions though.
Also it'd be good to triple check if you caught all n_neurons_dict to n_neurons renamings in case we stick to this decision.

openretina/cli/train.py
fededagos: Yes it will work, it was intentional :)

configs/data_io/hoefling_2024.yaml
fededagos: good point, will add

openretina/data_io/base.py
fededagos: I thought it was a bit cleaner to have in the data info. But I think I will revert it. `n_neurons_dict` appears in too many places, and renaming it everywhere will likely introduce some problems...

openretina/models/core_readout.py
fededagos: It's pytorch image models, I have added it in parentheses now so it is clear.

openretina/models/core_readout.py
```diff
@@ -104,16 +104,42 @@ def test_step(self, batch: tuple[str, DataPoint], batch_idx: int, dataloader_idx
         session_id, data_point = batch
         model_output = self.forward(data_point.inputs, session_id)
         loss = self.loss.forward(model_output, data_point.targets) / sum(model_output.shape)
-        correlation = -self.correlation_loss.forward(model_output, data_point.targets)
+        avg_correlation = -self.correlation_loss.forward(model_output, data_point.targets)
+        per_neuron_correlation = -self.correlation_loss._per_neuron_correlations
```
thomasZen: This is a bit inconsistent for two reasons:
- we negate the per neuron correlations, and the name doesn't change (in the line above it goes from correlation_loss to avg_correlation after the negation which makes more sense)
- we access a private member

So maybe directly store the negated result in the `correlation_loss` class and make it public by calling it `per_neuron_correlations`.

fededagos: I would want to keep it private as _per_neuron_correlations gets overwritten with every forward of the loss, storing the value of the last session essentially, so should only be accessed if people know what they are doing. (Also because if somebody wants explicitly the per_neuron loss only, the loss class offers an option to return only that already). I will resolve the naming inconsistency though

openretina/modules/losses/correlation.py
fededagos: I would leave it as a tensor so it can stay a buffer of the module. But do you know if I can `.cpu()` it without problems? i.e. if the rest of the module is on `cuda` and I put that buffer on cpu, will something break?

thomasZen: > But do you know if I can .cpu() it without problems? i.e. if the rest of the module is on cuda and I put that buffer on cpu, will something break?

Okay, good point. It's probably better to keep all on the same device.
Older stackoverflow comments say you can have different parameters on different devices, but I think that's also confusing: https://stackoverflow.com/questions/58926054/how-to-get-the-device-type-of-a-pytorch-module-conveniently

openretina/data_io/base.py
```diff
@@ -179,23 +179,49 @@ def normalize_train_test_movies(
...
+        stim_std = movies_dictionary.norm_std
+        input_shape = (
+            movies_dictionary.train.shape[0],
+            *movies_dictionary.train.shape[2:],
+        )
+    else:
+        norm_means = [movie.norm_mean for movie in movies_dictionary.values()]
+        norm_stds = [movie.norm_std for movie in movies_dictionary.values()]
+
+        if any(mean != norm_means[0] for mean in norm_means):
```
thomasZen, line 210: just as a not, it's also possible to write this check as:
```
if set(norm_means) > 1:
```
or to directly use a set above (but then indexing is harder an you'd have to write `next(iter(norm_means_set))` or so to get the value.

Anyway, just to show a potentially useful pattern, it's good as it is right now.

thomasZen (approved):
Looks good

fededagos:
@thomasZen I am merging now. Before running training for all models I will open a separate refactoring PR, that I will use in the end for that.

PR #159, opened 2025-10-09
Local Spike Triggered Average (LSTA) analysis

BaptisteLorenzi:
Toolkit + Notebook to compute LSTA vector fields as done in Goldin et al. 2022.
Example using Hoefling 2024 dataset (low res).


openretina/insilico/vector_field_analysis/vector_field_analysis.py
```diff
@@ -0,0 +1,615 @@
...
+import numpy as np
+import torch
+from git import Optional
+from PIL import Image
+from sklearn.decomposition import PCA
+
+from openretina.models.core_readout import BaseCoreReadout
+
+"""
+LSTA (Linear Spatio-Temporal Analysis) visualization toolkit.
```
thomasZen: I thought LSTA would mean Local Spike Triggered Average :sweat_smile: 
if there's already an explanation for it somewhere online (ie paper or abstract) feel free to link it here :)

openretina/insilico/vector_field_analysis/vector_field_analysis.py
```diff
@@ -0,0 +1,615 @@
...
+        KeyError: If session_id not found in model's normalization dictionary.
+    Notes
+    -------
+        - For single-channel models, uses session-specific normalization.
+        - For multi-channel models, uses default normalization.
+    """
+    if n_channels == 1:
+        movie_norm_dict_key = session_id
+    else:
+        movie_norm_dict_key = "default"
```
thomasZen: I know that I touched this to fix the mypy errors, but to me it seems a bit weird to decide the session id based on the number of channels?
Is that just working as the hoefling model has two channels, and the other one has multiple ones?

Another option would be to write here something like:
```
if session_id in model.data_info["movie_norm_dict"]
    movie_norm_dict_key = session_id
else:
    movie_norm_dict_key = "default"
```

or another solution if you have a better idea.

openretina/insilico/vector_field_analysis/vector_field_analysis.py
```diff
@@ -0,0 +1,615 @@
...
+
+    movie_norm_dict: dict[str, float] = model.data_info["movie_norm_dict"][movie_norm_dict_key]
+    stim_mean = movie_norm_dict["norm_mean"]
+    stim_std = movie_norm_dict["norm_std"]
+    for channel in range(n_channels):
+        movies[:, channel, :, :, :] = (movies[:, channel, :, :, :] - stim_mean) / stim_std
+    return movies
+
+
+def set_initial_frames_to_grey(movies, n_empty_frames):
```
thomasZen:
```suggestion
def set_initial_frames_to_grey(movies: np.ndarray, n_empty_frames: int) -> np.ndarray:
```

thomasZen: On second thought, I'm unsure about if this function is needed at all. I'd rather just delete it and move it inline:
- the function name is slightly confusing as it's not setting it to grey, but to the movie mean (this is written in the documentation, but people need to read it...)
- it's only one line of code and probably easier to write out that line instead of calling the function, then it's immediately clear what's happening.

openretina/insilico/vector_field_analysis/vector_field_analysis.py
```diff
@@ -0,0 +1,615 @@
...
+    n_channels = model.data_info["input_shape"][0]
+    target_h, target_w = model.data_info["input_shape"][1:3]
+
+    if image_library is not None and image_dir is not None:
+        raise ValueError("Provide either image_library or image_dir, not both.")
+    if image_dir is not None:
+        print(f"Loading images from {image_dir}...")
+        compressed_images = load_and_preprocess_images(image_dir, target_h, target_w, n_channels)
+    elif image_library is not None:
+        print("Using provided image library...")
```
thomasZen: If you want to keep the print statements, you can use a logger (that's what we decided for in other parts of the code, e.g., see [here](https://github.com/fededagos/open-retina/blob/84acac6fd0b8331ba04fc40f180d9981839fb183/openretina/models/core_readout.py#L22)). So at the beginning of the file write something like:
```
import logging
...
LOGGER = logging.getLogger(__name__)
```

and then you call the logger as follows:
```
LOGGER.info(f"Loading images from {image_dir}...")
```

thomasZen (approved):
Looks good, thanks for adding!

I made a few smaller suggestion (to remove and inline one function, and one in the notebook) which you can integrate if you agree.

notebooks/vector_field_analysis_hoefling_2024.ipynb
```diff
@@ -0,0 +1,477 @@
+{
```
thomasZen, line 1: <p>Maybe remove the <code>#</code> symbols here. I think it's okay to not execute it, but having it there as normal code seems better than as commented out code. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/open-retina/open-retina/pull/159/discussion/#comment-2419936973'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='26'/>

openretina/insilico/vector_field_analysis/vector_field_analysis.py
BaptisteLorenzi: This was just me being distracted.

openretina/insilico/vector_field_analysis/vector_field_analysis.py
BaptisteLorenzi: Yes, this is only because the exact dict path to the statistics of the training stimuli are not standardized yet. This is a little bit more elegant, but optimally this should be resolved in the future.

openretina/insilico/vector_field_analysis/vector_field_analysis.py
BaptisteLorenzi: Ok

notebooks/vector_field_analysis_hoefling_2024.ipynb
BaptisteLorenzi: Ok, it was just to have a smoother experience while pressing 'Run All'.

PR #161, opened 2025-10-30
Refactor readouts

fededagos:
Refactored all readout classes.
- Class inheritance is now consistent both for single and multi readouts
- Added back initialisation from mean activity of the neurons as the default option
- Readout names now refactored for clarity
- Improved (and new) readout documentations
- Changed configs to make karamanlis models causal
- Updated all hoefling models and added hoefling low res greyscale model
- Fixed spurious cache hits by checking full relative path instead of only the stem of the file_name


configs/logger/mlflow.yaml
```diff
@@ -1,8 +1,8 @@
 # MLflow logger configuration
 mlflow:
   _target_: lightning.pytorch.loggers.MLFlowLogger
-  experiment_name: ${exp_name}
-  tracking_uri: "${paths.log_dir}/openretina_assets/mlflow"
+  experiment_name: ${oc.env:MLFLOW_EXPERIMENT_NAME,${exp_name}}
+  tracking_uri: "${oc.env:MLFLOW_TRACKING_URI,${paths.log_dir}/openretina_assets/mlflow}"
```
thomasZen, line 7: This reads weirdly. I think it would be best to add a comment what is does above these two lines (something like check if the oc.env is present... (and say what oc.env is :) )

configs/model/core_gaussian_readout.yaml
```diff
@@ -32,12 +32,11 @@ readout:
...
+  gauss_type: "full"
   grid_mean_predictor: null
   shared_features: null
   init_grid: null
   shared_grid: null
-  mean_activity: null
 
 # training
-learning_rate: 0.001
+learning_rate: 0.01
```
thomasZen, line 42: was this on purpose, and did it improve training?

openretina/models/core_readout.py
```diff
@@ -34,15 +33,38 @@
...
+        Initializes a BaseCoreReadout module.
+
+        Args:
+            core (Core): The shared feature extraction core network.
+            readout (MultiReadoutBase): The multi-session readout module mapping core features to neuron outputs
+                per session.
+            learning_rate (float): Learning rate for network training.
+            loss (nn.Module, optional): Loss function for training. Defaults to PoissonLoss3d if None.
+            correlation_loss (nn.Module, optional): Loss used to compute correlation performance metric.
+                Defaults to CorrelationLoss3d (avg=True) if None.
```
thomasZen: Maybe this should be called validation_loss? At least say in the docstring that it's not used for training/calculation of gradients, but only for validation.

openretina/models/core_readout.py
```diff
@@ -227,6 +250,16 @@ def pretrained_cfg(self) -> dict[str, Any]:
 
 
 class UnifiedCoreReadout(BaseCoreReadout):
+    """
+    A highly flexible core-readout model for multi-session neural data, designed for Hydra config workflows.
```
```suggestion
    A flexible core-readout model for multi-session neural data, designed for Hydra config workflows.
```

openretina/models/core_readout.py
```diff
@@ -227,6 +250,16 @@ def pretrained_cfg(self) -> dict[str, Any]:
...
+
+    This class is the recommended entry point for defining core-readout models via config files using Hydra.
+    It allows unified instantiation of arbitrary core and readout modules, specified via DictConfig,
+    enabling rapid experimentation and extensibility. Supports all multi-session settings, custom core/readout
+    combinations, and seamless integration with configuration-driven pipelines.
```
thomasZen: by configuration driven pipelines do you just mean using hydra for training, or also things like hyperparameter tuning? 
In theory, it would be nice to just do a hyperparameter tuning across hyperparameters and core/readout combinations. But I don't know if that's actually doable now already.

configs/model/core_klindt_readout.yaml
```diff
@@ -19,14 +19,15 @@ core:
...
 readout:
-  _target_: openretina.modules.readout.multi_readout.MultiKlindtReadoutWrapper
+  _target_: openretina.modules.readout.multi_readout.MultiFactorizedReadout
   _convert_: object
   in_shape: ???
-  mask_l1_reg: 0.0
-  weights_l1_reg: 0.0
+  mask_l1_reg: 5.0
+  weights_l1_reg: 5.0
+  laplace_mask_reg: 1.0
```
thomasZen, line 30: This is for consistency I assume between different readout classes? I could also imagine that you'd get different hyperparameters for different readouts (e.g. @BaptisteLorenzi tuned them for a different model I think)

openretina/models/core_readout.py
```diff
@@ -227,16 +250,50 @@ def pretrained_cfg(self) -> dict[str, Any]:
...
+                Hydra config for instantiating the core module (should specify class and params).
+            readout (DictConfig):
+                Hydra config for the readout module (specifies type and custom session-aware params).
+            learning_rate (float, optional):
+                Learning rate for model training. Defaults to 0.001.
+            data_info (dict[str, Any], optional):
+                Additional metadata dictionary, e.g., with input shape and neuron mapping.
+        """
+        # force in_shape and hidden_channels to be a tuple, they can be a dictConfig due to hydra
+        # this lead to an error when logging hyperparameters with the csv logger during training
```
```suggestion
        # Make sure in_shape and hidden_channels are a tuple (with hydra configs they are a `omegaconf.listconfig.ListConfig`).
        # This lead to an error when logging hyperparameters with the csv logger during training.
```

openretina/models/core_readout.py
```diff
@@ -247,18 +304,34 @@ def __init__(
         if "in_shape" not in readout:
             in_shape_readout = self.compute_readout_input_shape(in_shape, core_module)
             readout["in_shape"] = (in_shape_readout[0],) + in_shape_readout[1:]
+
+        # Extract mean_activity_dict from data_info if available
+        mean_activity_dict = None
+        if data_info is not None and "mean_activity_dict" in data_info:
+            mean_activity_dict = data_info["mean_activity_dict"]
```
```suggestion
        mean_activity_dict = None if data_info is None else data_info.get("mean_activity_dict")
```
maybe shorten (`.get()` return None on default)

openretina/models/core_readout.py
```diff
@@ -247,18 +304,34 @@ def __init__(
...
+
         readout_module = hydra.utils.instantiate(
             readout,
             n_neurons_dict=n_neurons_dict,
+            mean_activity_dict=mean_activity_dict,
         )
 
-        # if calling save_hyperparameters after __init__ it leads to errors related to data_info["session_kwargs"]
```
thomasZen, line 255: It's also possible to only store selected attributes by passing a dict to save_hyperparameters.
I don't know how it looks in wandb/tensorflow, but the csv file also contained binary data which we could avoid, too.

openretina/models/core_readout.py
```diff
@@ -290,6 +363,14 @@ def __init__(
         color_squashing_weights: tuple[float, ...] | None = None,
         data_info: dict[str, Any] | None = None,
     ):
+        warnings.warn(
```
thomasZen, line 374:
```suggestion
        LOGGER.warning(
```

Maybe we should stick to the logger object for consistency (I don't know the warnings module well though, maybe it has it's purpose on top of the logging based LOGGER)

openretina/modules/readout/factorized_gaussian.py
```diff
@@ -169,3 +215,8 @@ def save_weight_visualizations(self, folder_path: str, file_format: str = "jpg",
             fig_axes_tuple[0].savefig(plot_path, bbox_inches="tight", facecolor="w", dpi=300)
             fig_axes_tuple[0].clf()
             plt.close()
+
+
+# Alias for backward compatibility
+class SimpleSpatialXFeature3d(GaussianMaskReadout):
+    pass
```
thomasZen: If the bias_param to bias change breaks backwards compatibility, should we just delete this?

openretina/modules/readout/multi_readout.py
```diff
@@ -58,50 +110,101 @@ def add_sessions(self, n_neurons_dict: dict[str, int]) -> None:
...
-        feature_loss = self[data_key].feature_l1(average=self.readout_reg_avg) * self.gamma_readout
-        mask_loss = self[data_key].mask_l1(average=self.readout_reg_avg) * self.gamma_masks
-        return feature_loss + mask_loss
+    def initialize(self, mean_activity_dict: dict[str, Float[torch.Tensor, " neurons"]] | None = None):
+        for data_key, readout in self.items():
+            mean_activity = mean_activity_dict[data_key] if mean_activity_dict is not None else None
+            readout.initialize(mean_activity)
+
+    def regularizer(self, data_key: str | None = None, reduction: Literal["sum", "mean"] = "sum"):
+        reduction = "mean" if self.readout_reg_avg else reduction
```
thomasZen: I against this line, can't we just delete it and let the caller pass if he wants the "sum" or the "mean".
I think it's confusing if you call `regularizer(reduction="sum")` but then it uses "mean" because `self.readout_reg_avg` is True.

openretina/modules/readout/multi_readout.py
```diff
@@ -3,51 +3,103 @@
...
+        clone_readout (bool): whether to clone the first data_key's readout to all other readouts, only allowing for a
+                                scale and offset. This is a rather simple method to enforce parameter-sharing
+                                between readouts.
+
+        gamma_readout (float): regularization strength
+
+        **kwargs: additional keyword arguments to be passed to the base_readout's constructor
     """
 
+    _base_readout_cls: type[Readout] | None = None
```
thomasZen, line 45: I think you  can actually enforce this in the type hierarchy like this (ABC stands for abstract base class):
```
from abc import ABC, abstractmethod
from typing import Type

class MultiReadoutBase(nn.ModuleDict, ABC):
    @property
    @abstractmethod
    def base_readout_cls(self) -> Type[Readout]:
        """Concrete subclasses must return a Readout type."""
        raise NotImplementedError

class MultiGaussianReadout(MultiReadoutBase):
    _base_readout_cls: ClassVar[Type[Readout]] = GaussianFactorizedReadout

    @property
    def base_readout_cls(self) -> Type[Readout]:
        return GaussianFactorizedReadout
```

thomasZen: then we'd avoid the if None check.

openretina/modules/readout/multi_readout.py
```diff
@@ -58,50 +110,101 @@ def add_sessions(self, n_neurons_dict: dict[str, int]) -> None:
...
+    def forward(self, *args, data_key: str | None = None, **kwargs) -> torch.Tensor:
+        if data_key is None:
+            readout_responses = []
+            for readout_key in self.readout_keys():
+                resp = self[readout_key](*args, **kwargs)
+                readout_responses.append(resp)
+            response = torch.concatenate(readout_responses, dim=-1)
+        else:
+            response = self[data_key](*args, **kwargs)
+        return response
```
thomasZen: can't this just be the standard forward function, I think it's very similar between multi_readout classes.

thomasZen: I see that it's sometimes slightly different though (e.g. it uses a nonlinearity, and rearranges inputs).

thomasZen (commented):
Awesome, I think we have to be careful communicating that we broke backwards compatibility for some older models, though (e.g. we should mention it in slack and ask if this causes issues).

I added a few suggestions etc.

configs/model/core_gaussian_readout.yaml
fededagos: Yes it was on purpose. It both improved and it is consistent with the other model configs

openretina/models/core_readout.py
fededagos: I think we should rename it yes

openretina/models/core_readout.py
fededagos: yeah I kind of meant both. I think with the current setup one could already do what you describe, it would only be a bit convoluted to specify

configs/model/core_klindt_readout.yaml
fededagos: Yes, also because I think these readout are meant to be sparse, so I did not want the default to be no sparsity at all

openretina/models/core_readout.py
fededagos: It is indeed a bit messy right now. I will double check again what gets saved and exclude more things at least from the logging.

openretina/models/core_readout.py
fededagos: warnings get printed even without a logger, but loggers do not show if a logger is not set up (i.e. in jupyter). There is a way to have warnings shown as logs though if a logger is set up, I did not know this until checking this out now:
https://docs.python.org/3/library/logging.html#logging.captureWarnings
I will maybe add this flag to the global logger in training.

openretina/modules/readout/factorized_gaussian.py
fededagos: That actually makes sense

openretina/modules/readout/multi_readout.py
fededagos: I agree that it is a bit dirty... The issue though is that a lot of readouts still expect this reg_avg flag... I will look at a cleaner way to do this

openretina/modules/readout/multi_readout.py
fededagos: Very good suggestion, I will look into it

openretina/modules/readout/multi_readout.py
fededagos: Yes exactly. But you are right that at least two readouts use the more standard one. I can try to unify those that do

openretina/models/core_readout.py
fededagos: I have renamed it package-wide

openretina/models/core_readout.py
fededagos: I did that :)

openretina/modules/readout/multi_readout.py
fededagos: I was trying to implement this and it turned out to be a bit more code (and in my opinion a bit less readable), then this version

openretina/modules/readout/multi_readout.py
fededagos: Did it :)

thomasZen (approved):
Still looks good, thanks for the edits.

PR #166, opened 2025-12-05
Karamanlis models trained on mouse and on marmoset data

thomasZen:
Additionally fixed a bug (and added tests) for which the hoefling grey_scale model was returned when requesting the normal hoefling model in case the grey_scale model was downloaded first (those models had common prefixes on huggingface).


fededagos (approved):
Looks good! It is great to have a test for this so we can avoid weird behaviour in the future :)

PR #167, opened 2025-12-05
Integrate LNP into core-readout structure

thomasZen:
There's still the SingleCellSeparatedLNP, we can integrate that at a later point.

I also added each model training config into the tests and divided the end-to-end test into jupyter tests and model training tests (and fixed the GRU core model config while doing that).

It might be that github got more stringent with storage, as the jupyter test fails when downloading the high res movie due to no space left on device.

## Results Hoefling low res
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃        Test metric        ┃       DataLoader 0        ┃       DataLoader 1        ┃       DataLoader 2        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     test_correlation      │    0.10988865792751312    │    0.0995618924498558     │    0.14662180840969086    │
│         test_loss         │    1485.9398193359375     │     640.0553588867188     │     58.63211441040039     │
└───────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────┘
```

## Results Hoefling low res with bias
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃        Test metric        ┃       DataLoader 0        ┃       DataLoader 1        ┃       DataLoader 2        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     test_correlation      │    0.15619322657585144    │    0.08392297476530075    │    0.26656538248062134    │
│         test_loss         │     1027.402587890625     │       399.169921875       │    37.101070404052734     │
└───────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────┘

## Results on maheswaranathan_2023
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃        Test metric        ┃       DataLoader 0        ┃       DataLoader 1        ┃       DataLoader 2        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     test_correlation      │    0.04313581809401512    │   0.025508105754852295    │    0.08250889405306212    │
│         test_loss         │     473.0135498046875     │     312.7330627441406     │    24.196617608142617     │
└───────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────┘
```

## Results on maheswaranathan_2023 before this PR
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃        Test metric        ┃       DataLoader 0        ┃       DataLoader 1        ┃       DataLoader 2        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     test_correlation      │    0.04313581809401512    │   0.025508105754852295    │    0.08250889405306212    │
│         test_loss         │     473.0135498046875     │     312.7330627441406     │    24.196617608142617     │
└───────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────┘
```


openretina/models/core_readout.py
```diff
@@ -302,10 +306,11 @@ def __init__(
         # Make sure in_shape and hidden_channels are a tuple
         # (with hydra configs they can be a `omegaconf.listconfig.ListConfig`).
         # This lead to an error when logging hyperparameters with the csv logger during training.
-        hidden_channels = tuple(hidden_channels)
         in_shape = tuple(in_shape)
+        if hidden_channels is not None:
```
fededagos, line 313: This is good, I was running into similar problems for the transformer core, I might have resolved it in a more dirty way there...

openretina/modules/readout/linear_nonlinear_poison.py
```diff
@@ -0,0 +1,79 @@
...
+import torch.nn.functional as F
+from einops import rearrange
+from jaxtyping import Float, Int
+
+from openretina.modules.layers import regularizers
+from openretina.modules.readout.multi_readout import Readout
+
+
+class LNPReadout(Readout):
+    # Linear nonlinear Poisson
```
fededagos, line 16: Could you add a bit more documentation, either here or for the multi one (or both)? I think in this case it might be particularly helpful, as this is kind of readout is a bit special in that it is always intended to be used with a dummycore to act as intended

fededagos (approved):
Is is nice to have training test for all combinations, will be much easier to tell when things break from now on. I would only include a bit more docs for the LNP readout, otherwise looks great. 
Not sure about the notebook error though, it seems weird that it is due to space. Maybe just re-running the task will resolve?

fededagos:
> Is is nice to have training test for all combinations, will be much easier to tell when things break from now on. I would only include a bit more docs for the LNP readout, otherwise looks great. Not sure about the notebook error though, it seems weird that it is due to space. Maybe just re-running the task will resolve?

I actually realised that in the notebooks test the data notebook downloads the maheswaranathan dataset, which is also quite heavy. Maybe after each notebook is run in the task we could clean up the openretina cache directory? Not sure how to do that just in the workflow though

openretina/modules/readout/linear_nonlinear_poison.py
thomasZen: I added a bit more documentation for both.

thomasZen:
> Maybe after each notebook is run in the task we could clean up the openretina cache directory? Not sure how to do that just in the workflow though

I didn't figure out how to do that after each individual notebook in an easy way. I changed it now to first run the disk space heavy data notebook, clean the cache, and then run the remaining notebooks. (maybe we find a better solution in the future, but I think that's reasonable for now)

PR #168, opened 2025-12-11
Add script for evaluation of models

thomasZen:
- Changed temporal_kernel_sizes of high res cnn model to make sure that the frames that get cut by temporal padding are the same compared to the low res model (before that 40 frames were cut, now it's 30 frames)
- Add openretina eval script


openretina/cli/eval.py
```diff
@@ -0,0 +1,140 @@
...
+            model_responses_torch = model.forward(movies, data_key=session)
+            targets = dataset.responses.to(device).unsqueeze(0)
+            poisson_loss_session = poisson_loss(model_responses_torch, targets)
+
+        results["poisson_loss"].append(poisson_loss_session.cpu().numpy())
+
+        model_responses = model_responses_torch.squeeze(0).cpu().numpy()
+        model_responses_dict[session] = model_responses
+
+        responses_by_trial = dataset.test_responses_by_trial.cpu().numpy()
```
fededagos: I think right now this might not be implemented for some datasets, or at least it could be implemented but it is not right now. I think I would like to have another pass at datasets and add this for all before we assume we always have that information, and if we do not deal with that here somehow

thomasZen: Good point, I'll wrap this in a try/catch blog and fall back to the average responses.

fededagos (approved):
Approving and merging for now.
@thomasZen please have a look at my changes once you are back :)

PR #169, opened 2025-12-12
Dev: allow loading of legacy models that use bias_param instead of bias_param in readout

thomasZen:
To make the following work:
```
from openretina.models.core_readout import UnifiedCoreReadout, load_core_readout_model
load_core_readout_model("https://huggingface.co/datasets/open-retina/open-retina/blob/main/model_checkpoints/24-01-2025/hoefling_2024_base_low_res.ckpt", device="cpu")
```

### DIFF
--- a/openretina/models/core_readout.py
+++ b/openretina/models/core_readout.py
@@ -425,6 +425,7 @@ class ExampleCoreReadout(BaseCoreReadout):
         super().__init__(core=core, readout=readout, learning_rate=learning_rate, data_info=data_info)
 
     def on_load_checkpoint(self, checkpoint) -> None:
+    """ To support legacy models that use `bias_param` instead of `bias` in their readout layers.
         state_dict = checkpoint["state_dict"]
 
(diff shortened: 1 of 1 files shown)


openretina/models/core_readout.py
```diff
@@ -424,6 +424,16 @@ def __init__(
 
         super().__init__(core=core, readout=readout, learning_rate=learning_rate, data_info=data_info)
 
+    def on_load_checkpoint(self, checkpoint) -> None:
```
thomasZen, line 427:
```suggestion
    def on_load_checkpoint(self, checkpoint) -> None:
    """ To support legacy models that use `bias_param` instead of `bias` in their readout layers.
```

openretina/models/core_readout.py
```diff
@@ -424,6 +424,17 @@ def __init__(
...
+    """ To support legacy models that use `bias_param` instead of `bias` in their readout layers.
```
```suggestion
    """ To support legacy models that use `bias_param` instead of `bias` in their readout layers. """
```

fededagos (commented):
Nice! Quite a clean solution as well, I did not know lightning supported this.

fededagos (approved):

PR #170, opened 2026-01-02
Training enhancements

fededagos:
Miscellaneous changes to enhance flexibility during training:
- Support passing different losses and validation losses to the trainer
- Support custom scheduler
- Support custom optimizer


configs/lr_scheduler/reduce_lr_on_plateau.yaml
```diff
@@ -0,0 +1,19 @@
+# ReduceLROnPlateau scheduler configuration
+# This is the default scheduler used in the codebase
```
thomasZen: I'm wondering if we should keep this comment?
Do we plan to keep this scheduler as the default for long, let's say the next two years?

configs/optimizer/sgd.yaml
```diff
@@ -0,0 +1,7 @@
+# SGD (Stochastic Gradient Descent) optimizer configuration
+_target_: torch.optim.SGD
+lr: ${model.learning_rate}
+momentum: 0.9
+dampening: 0.0
+weight_decay: 0.0
+nesterov: false
```
thomasZen, line 7: Do we need all of these? Maybe just keep two, e.g. sgd and adamw (adamw is strictly better than adam I think) and then people can write their own config if they want to use something non standard.

configs/karamanlis_2024_core_readout.yaml
```diff
@@ -2,13 +2,15 @@ defaults:
   - data_io: karamanlis_2024
   - dataloader: karamanlis_2024
   - model: base_core_readout
+  - optimizer: adamw
+  - lr_scheduler: reduce_lr_on_plateau
   - training_callbacks:
-    - early_stopping
-    - lr_monitor
-    - model_checkpoint
+      - early_stopping
```
thomasZen, line 8: out of interest, was this change manual, some codestyle change or done by an LLM?
both is fine, if it's codestyle I'd double check my IDE settings, though.

openretina/utils/optimizer_utils.py
```diff
@@ -0,0 +1,251 @@
...
+    )
+
+    return {
+        "scheduler": scheduler,
+        "monitor": "val_correlation",
+        "frequency": 1,
+    }
+
+
+def _handle_special_scheduler_cases(
```
thomasZen, line 159: Unsure if this is it's own function. It's basically 5 lines if you inline it:
```
target = scheduler_config.get("_target_", "")
if "OneCycleLR" in target:
   _handle_one_cycle()
elif "ReduceLROnPlateau" in target:
   _handle_plateau()
```

thomasZen (approved):
Nice, I'm also fine with the implementation but maybe one iteration to reduce the code footprint could make sense. The parts which I do not like too much (but feel free to disagree and merge without addressing):
- optimizer configs for pytorch optimizers that we do not heavily use
- many small functions in optimizer_utils
- defaults for optimizers/lr_schedulers in the code when passing None: they are defined in sub-configs already, I think it's better to explicitly have them written out in the config by people

configs/karamanlis_2024_core_readout.yaml
fededagos: I added a new .yaml formatting extension is cursor :)

openretina/utils/optimizer_utils.py
fededagos: You are right, I can simplify this

configs/optimizer/sgd.yaml
fededagos: Fair point, will get rid of RMSProp and adam

openretina/utils/optimizer_utils.py
fededagos: Actually I think I might keep it, in case we add some more special treatments in the future

configs/lr_scheduler/reduce_lr_on_plateau.yaml
fededagos: True, changed this

PR #171, opened 2026-01-14
Add Palmer 2024 dataset loading functionality

fededagos:
For some reason, training does not work at all (neither train nor val loss increase):
This should be investigated more in depth. It suggests a mismatch between stims and responses that I have not been able to solve so far.


thomasZen:
AI assessment:

• The evidence points to a target-construction bug, not a gross stimulus/response ordering mismatch.

  ### Main findings

  - The repeat loader trains on binary 0/1 responses averaging 0.00994, while validation and model initialization use firing_rate_60ms, averaging 0.0543 and 0.1510, respectively.
  - Across the entire HDF5 file, firing_rate_60ms == binned × 16.6667 exactly. Thus training immediately drives predictions roughly 15× below their initialization, away from validation’s units.
  - The repeat configuration passes response_key (configs/dataloader/palmer_2024_repeats.yaml:8), but repeats_dataloaders does not accept it (openretina/data_io/palmer_2024/dataloaders.py:177). The configured
    path raises TypeError before training.

  - Averaged HDF5 responses include all 91 repeat slots, including padded zeros. Because valid repeat counts are 80–91, four movies’ averages are underestimated by 6.6–12.1%.
  - The repeat training dataset includes all four movies, including the movie subsequently selected for validation (openretina/data_io/palmer_2024/dataloaders.py:253, openretina/data_io/palmer_2024/
    dataloaders.py:271). This is validation leakage.

  - Even the standard loader has a strong scene-distribution shift: with seed 42, training targets average 0.1833, versus 0.0543 for the held-out movie. Validation Poisson loss therefore need not follow training
    loss, even with correct alignment.

  ### Alignment assessment

  I found no evidence of a movie-order mismatch:

  - Every precomputed binned response equals the mean of that same movie’s repeat tensor exactly.
  - Movie indices and names are used consistently by both stimulus and response loaders.
  - Response split-half reliability averages 0.669; cross-movie correlations average only 0.010.
  - A simple spatiotemporal ridge baseline obtained correlation 0.096 on correctly paired data, versus 0.031 with the movie identities rotated and 0.057 with a 200-frame temporal shift.

  The 1,203-response/1,200-frame discrepancy remains worth documenting, but the final response bin is entirely zero and the last three bins have very low activity, supporting the interpretation that they are
  trailing acquisition bins. It does not look like the principal failure.

  The source dataset also says the first 500 ms—30 frames—should be excluded, which the PR does not currently do. Dryad documentation

  ### Recommended correction order

  1. Use one response representation throughout. The simplest option is binned for training, validation, testing, and bias initialization.
  2. Recompute averages using only nreps[movie] valid trials.
  3. Remove the validation movie from repeat training.
  4. Fix the response_key signature/config mismatch.
  5. Exclude the first 30 frames of every movie.
  6. Add an overfit-one-chunk test plus shifted/movie-permuted alignment controls.

thomasZen:
I got this now for correlations:
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃        Test metric        ┃       Train               ┃       Validation          ┃       Test                ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     CorrelationLoss3d     │    0.2250608652830124     │    0.2953479290008545     │    0.03275247663259506    │
```
It also trains on the averaged responses now, which should be faster but in the limit the same as training on individual repeats with large batch sizes. Training  converges, but test correlation is still bad.

### AI thinks
• The original 2024 study did not divide movies into train, validation, and test sets.

  Instead, it:

  - Used all five movies: tree, water, grasses, fish, and self-motion.
  - Fit a separate time-dependent maximum-entropy model for each movie.
  - Validated each model by fitting it on half of that movie’s repeated trials and testing pairwise neural statistics on the remaining half.
  - Compared the fitted interaction parameters across movies; it did not train on some movies and predict firing rates for an unseen movie. [Original PNAS study](https://pubmed.ncbi.nlm.nih.gov/39700141/)

  The paper actually reports that complete movie-specific models did not transfer well between movies, although the strongest cell-to-cell coupling structure was consistent. So the current OpenRetina task—train
  on tree/grasses/optic flow and predict fish—is substantially harder and is not a reproduction of the original modeling setup.

PR #172, opened 2026-01-14
Feature/ln gaussian fits

Mvystrcilova:
I added a gaussian fitting option to the single cell LNPs.


openretina/models/linear_nonlinear.py
```diff
@@ -66,6 +67,9 @@ class SingleCellSeparatedLNP(LightningModule):
     laplace_padding:
         Passed through to regularizer constructors as `padding=...`. For GaussianLaplaceL2,
         a `kernel=...` argument is also supplied.
+    fig_gaussian:
```
fededagos, line 70: Small typo :)

openretina/utils/model_utils.py
```diff
@@ -78,3 +78,42 @@ def get_core_output_based_on_dimensions(model_config):
...
+    kernel_size,
+    amplitude,
+    lower=None,
+    upper=None,
+):
+    if lower is not None:
+        amplitude.data.clamp_(lower, upper)
+
+    covariance_matrix = get_cholesky_covariance(l11_raw, l22_raw, l21)
+    inverse_covariance_matrix = torch.inverse(covariance_matrix)
```
fededagos, line 110: If you want to use the Cholesky factor in here directly (instead of the computed cov), apparently `torch.cholesky_inverse` should be more numerically stable (according to Claude), and you avoid an outer product

openretina/models/linear_nonlinear.py
```diff
@@ -283,12 +309,56 @@ def normalize_kernels(self):
...
+        self.mean_y = nn.Parameter(torch.zeros(1), requires_grad=True)
+        self.l11_raw = nn.Parameter(torch.ones(1) * l11, requires_grad=True)
+        self.l22_raw = nn.Parameter(torch.ones(1) * l22, requires_grad=True)
+        self.l12 = nn.Parameter(torch.zeros(1), requires_grad=True)
+
+        x_coord = torch.linspace(-kernel_size[0] // 2, kernel_size[0] // 2, kernel_size[0])
+        y_coord = torch.linspace(-kernel_size[1] // 2, kernel_size[1] // 2, kernel_size[1])
+        x, y = torch.meshgrid(x_coord, y_coord)
+
+        self.register_buffer("x_coord", x_coord.clone())
```
fededagos, line 343: I believe these (`x_coord` and `y_coord`) are registered but never used explicitly. Is this on purpose?

openretina/utils/model_utils.py
```diff
@@ -78,3 +78,42 @@ def get_core_output_based_on_dimensions(model_config):
...
+    y,
+    mean_x,
+    mean_y,
+    kernel_size,
+    amplitude,
+    lower=None,
+    upper=None,
+):
+    if lower is not None:
+        amplitude.data.clamp_(lower, upper)
```
fededagos, line 107: Amplitude seems to be passed as a default float and not wired in from anywhere, so this will break. Was amplitude supposed to be a learnable param as well?

fededagos (commented):
Looks good overall, thanks for this! There are a couple comments maybe worth addressing, and I would ask you if you can please clear all outputs from the notebooks, but then it is good to merge!

PR #173, opened 2026-02-02
Dev: hyperparameter tuning continues after errors

thomasZen:
This requires to define objective_direction in the config and set it to `maximize` or `minimize`.
It will still log the full exception then and assign the run +-float('inf') as a score, and continue with the remaining hyperparameters.

Spurious: removed `from git import Optional`


fededagos (approved):
Looks great, merging immediately :)

PR #174, opened 2026-02-04
Add dataset statistics to evaluation loop

fededagos:


openretina/utils/eval_utils.py
```diff
@@ -0,0 +1,248 @@
+"""Utilities for model evaluation and result aggregation."""
+
+from __future__ import annotations
```
thomasZen: I think you don't need that anymore:
> from __future__ import annotations ([PEP 563](https://peps.python.org/pep-0563/)) will continue to exist with its current behavior at least until Python 3.13 reaches its end-of-life. Subsequently, it will be deprecated and eventually removed.

see: https://peps.python.org/pep-0749/

openretina/utils/eval_utils.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Model identification
+    model_path: str
+    model_tag: str
+    dataset_name: str
+    species: str | None
+    data_split: str
+
+    # Model characteristics
+    lag: int
```
thomasZen:
```suggestion
    temporal_lag: int
```
I think this is clearer for people who read the evaluation report

openretina/utils/eval_utils.py
```diff
@@ -0,0 +1,248 @@
...
+            i += 1
+
+        return cls(
+            model_path=model_path,
+            model_tag=model_tag,
+            dataset_name=dataset_name,
+            species=species,
+            data_split=data_split,
+            lag=lag,
+            n_test_repeats_min=min(n_trials_per_session) if n_trials_per_session else 0,
```
thomasZen:
```suggestion
            n_test_repeats_min=min(n_trials_per_session, default=0.0),
```

openretina/utils/eval_utils.py
```diff
@@ -0,0 +1,248 @@
...
+            n_test_repeats_max=max(n_trials_per_session) if n_trials_per_session else 0,
```
thomasZen:
```suggestion
            n_test_repeats_max=max(n_trials_per_session, default=0),
```

thomasZen (commented):
already some (nitpick) feedback on openretina/utils/eval_utils.py, but looks good so far.

configs/dataloader/hoefling_2024.yaml
```diff
@@ -2,5 +2,6 @@ _target_: openretina.data_io.hoefling_2024.dataloaders.natmov_dataloaders_v2
 _convert_: object # Keep python objects their class and does not convert to DictConfig
 batch_size: 64
 train_chunk_size: 50
+clip_length: 150 # From hoefling_2024.constants.CLIP_LENGTH
```
thomasZen:
```suggestion
clip_length: 150
```
it might actually be good to use this config as ground truth, not the hard coded values

configs/vystrcilova_2024_nm_cnn.yaml
```diff
@@ -23,9 +25,9 @@ dataloader:
 model:
   in_shape:
     - ${data_io.input_channels}
-    - ${data_io.placeholder_time_bins}
-    - ${data_io.placeholder_height}
-    - ${data_io.placeholder_width}
+    - 132 # time bins
```
thomasZen: that should move to the data_io config, right?

configs/vystrcilova_2024_wn_cnn.yaml
```diff
@@ -23,9 +25,9 @@ dataloader:
 model:
   in_shape:
     - ${data_io.input_channels}
-    - ${data_io.placeholder_time_bins}
-    - ${data_io.placeholder_height}
-    - ${data_io.placeholder_width}
+    - 132 # time bins
```
thomasZen: see above

configs/data_io/sridhar_nm_2025.yaml
```diff
@@ -20,12 +19,13 @@ excluded_cells: ${dataloader.excluded_cells}
...
 stimuli:
-  _target_: openretina.data_io.sridhar_2025.stimuli.build_placeholder_movies
-  session_ids: ${data_io.sessions}
+  _target_: openretina.data_io.sridhar_2025.stimuli.build_movies_from_responses
+  base_path: ${paths.data_dir}
+  response_files: ${data_io.response_files}
   channels: ${data_io.input_channels}
-  height: ${data_io.placeholder_height}
-  width: ${data_io.placeholder_width}
-  time_bins: ${data_io.placeholder_time_bins}
```
thomasZen, line 28: maybe keep time bins, if there's no good reason to remove it

openretina/cli/eval.py
```diff
@@ -228,99 +230,61 @@ def evaluate_model(cfg: DictConfig) -> float:
...
-    else:
-        print(f"Variance ratio cutoff: {var_ratio_cutoff} (NOT APPLIED - no trial/repeats data available)")
-        print("Note: var_ratio could not be computed (no trial/repeats data). All neurons included.")
-    print("-" * 80)
+    # Compute dataset statistics by iterating over the actual dataloaders
+    dataset_stats = compute_dataloader_statistics(dataloaders)
+
+    # Extract metadata from config
+    model_tag = cfg.evaluation.get("model_tag", cfg.evaluation.model_path)
+    dataset_name = cfg.get("exp_name", "unknown")
```
thomasZen: That's not the dataset name, but the experiment name, right?

openretina/data_io/sridhar_2025/stimuli.py
```diff
@@ -86,3 +90,78 @@ def build_placeholder_movies(
             norm_std=norm_std,
         )
     return movies
```
thomasZen, line 92: Can we remove this then?

thomasZen: or does it still get used during training?

openretina/data_io/base_dataloader.py
```diff
@@ -606,3 +609,158 @@ def multiple_movies_dataloaders(
             )
 
     return dataloaders
+
+
+FrameFingerprint = tuple[float, float, float, float, float]
+TransitionFingerprint = tuple[FrameFingerprint, FrameFingerprint]
+
+
+def _compute_frame_fingerprints(inputs: torch.Tensor) -> list[FrameFingerprint]:
```
thomasZen: maybe this could go in a dedicated file (e.g. frame_fingerprints), as it's not really essential to dataloading and base_dataloader already has 500+ lines

thomasZen (approved):
Looks reasonable, added smaller comments, though.

Also, does this always run during evaluation now? That also means that we'd always load the whole training set during evaluation, which is maybe not what we want.

openretina/cli/eval.py
fededagos: There is usually some overlap, but you are right. Will fix :)

configs/vystrcilova_2024_nm_cnn.yaml
fededagos: yeah, moved

PR #176, opened 2026-02-19
Fix goldin 2022 core readout config

thomasZen:
Changes:
- Renamed logged value from `val_correlation` to `val_validation_metric` as validation_loss can be different from correlation (and update configs to reflect that). Also moved the negation into the correlation related classes, instead of also automatically negating the result of other losses like the poisson loss.
- Print the names of the metrics at the end of training during the testing step
- During creating test snippets, adhere to clip_length if not allow_over_boundaries
- Update goldin configs to reflect above
- Fix discrepancy between test (contained spiking rate in Hz) and test by trial (contained number of spikes) in goldin responses


configs/dataloader/goldin_2022.yaml
fededagos, line 1: Can you add a comment here, and maybe make a docs entry as well, about why this should be always used for image models especially? It links to the changes you did downstream about testing ranges.

configs/lr_scheduler/reduce_lr_on_plateau.yaml
```diff
@@ -12,7 +12,7 @@ cooldown: 0
 eps: 1.0e-8
 
 # PyTorch Lightning specific parameters
-monitor: val_correlation # Metric to monitor
+monitor: val_validation_metric # Metric to monitor
```
fededagos: I am adding the comment here, but it applies everywhere. It is a bit of a semantic nitpick, but I think we should not use "validation_metric" but "evaluation_metric" everywhere in these changes. That is so that we do not confound the use of validation as in "validation split" with validation as in "way to validate (i.e. _evaluate_) a model. So this could become val_evaluation_metric. Also, tbf, we could just use "evaluation_metric" or "eval_metric".

configs/goldin_2022_core_readout.yaml
```diff
@@ -40,8 +42,12 @@ model:
     gamma_input: 0.4654576767872573
     convolution_type: "time_independent"
   readout:
-    mask_l1_reg: 5.924553274051563
-    feature_weights_l1_reg: 6.075840974162483
+    mask_l1_reg: 0.0  # 5.924553274051563
+    feature_weights_l1_reg: 0.0  # 6.075840974162483
+    positive: false
+  validation_loss:
```
fededagos: Again, I think this could also become "evaluation_loss" for clarity

openretina/data_io/base_dataloader.py
```diff
@@ -253,41 +253,6 @@ def get_next_bound(value, bounds):
     return shifted_indices
 
 
-def handle_missing_start_indices(
```
fededagos, line 256: Agree that this was a bit redundant code

openretina/models/core_readout.py
```diff
@@ -67,7 +67,7 @@ def __init__(
                 per session.
             learning_rate (float): Learning rate for network training.
             loss (nn.Module, optional): Loss function for training. Defaults to PoissonLoss3d if None.
-            validation_loss (nn.Module, optional): Loss used to compute correlation performance metric.
+            validation_metric (nn.Module, optional): Metric used to compute evaluate the model.
```
fededagos: Again, my preferred naming would be "evaluation_metric" or "eval_metric". You also used eval in the description here :) 
Btw, in the docstring I would also say that it can be a loss as well, it will just not get gradients

openretina/models/core_readout.py
```diff
@@ -497,6 +504,8 @@ def on_load_checkpoint(self, checkpoint) -> None:
         if len(readout_bias_keys) > 0:
             LOGGER.warning(f"Renamed the following readout bias keys: {readout_bias_keys}")
 
+        breakpoint()
```
fededagos: Forgot this :)

fededagos (commented):
Great to have the Goldin model fixed, and with it compatibility with image models in general!
The main comment that I would push to solve is changing all instances of validation loss/metric etc. with "evaluation" or "eval", see main comments. I do think that it is a very nice step to finally generalising this aspect as well

configs/lr_scheduler/reduce_lr_on_plateau.yaml
thomasZen: Good suggestion, I was struggling with these, too, but did not come up with something meaningful.

So for the logged metrics that's mean:
- `train_evaluation_metric`
- `val_evaluation_metric`

And we would change the class and the configs to take an `evaluation_loss`.

thomasZen: The downside is that we would invalidate old configs. I'll check for a bit if it's easy to do still support configs that talk about `validation_loss`, but would go ahead in any case.

openretina/models/core_readout.py
thomasZen: uups, it's gone now :)

configs/goldin_2022_core_readout.yaml
thomasZen: done

openretina/models/core_readout.py
thomasZen: done

configs/goldin_2022_core_readout.yaml
fededagos: I still see "validation" in the diff 😬

openretina/models/linear_nonlinear.py
```diff
@@ -103,12 +103,12 @@ class SingleCellSeparatedLNP(LightningModule):
         - regularization_loss_core
         - train_total_loss
         - train_loss
-        - train_correlation
+        - train_validation_loss
```
fededagos: nitpick because it'a doc, but I guess it should be train_evaluation_loss

openretina/models/linear_nonlinear.py
```diff
@@ -103,12 +103,12 @@ class SingleCellSeparatedLNP(LightningModule):
...
     Validation:
         - val_loss
         - val_regularization_loss
         - val_total_loss
-        - val_correlation
+        - evaluation_loss
```
fededagos: val_evaluation_loss?

fededagos (approved):
Looks good! A couple teeny tiny things to double check, but should be okay to merge otherwise :)

PR #179, opened 2026-02-26
Dev: Potential bugfixes and small improvements

thomasZen:
Bugs found with AI, but fixes implemented manually.


openretina/cli/train.py
```diff
@@ -93,7 +93,7 @@ def train_model(cfg: DictConfig) -> float | None:
                     data_info[k] = model.data_info[k] | data_info[k]
                 else:
                     log.warning(f"data_info key {k} when loading model: {model.data_info.keys()=}")
-            data_info["n_neurons_dict"] = model.data_info
+            data_info["n_neurons_dict"] = model.data_info["n_neurons_dict"]
```
thomasZen, line 96: this seemed to be obviously false as it assigns the whole data_info dict to `data_info["n_neurons_dict"]`.

openretina/data_io/hoefling_2024/stimuli.py
```diff
@@ -119,11 +119,11 @@ def gen_start_indices(
     num_train_clips = num_clips - len(val_clip_idx)
 
     if random_sequences.shape[1] == 1:
-        start_idx_dict["train"] = list(np.arange(0, clip_length * (num_train_clips - 1), chunk_size, dtype=int))
+        start_idx_dict["train"] = list(np.arange(0, clip_length * num_train_clips, chunk_size, dtype=int))
```
thomasZen, line 122: AI complained about this, and I think it's right.

We currently seem to not use the last train clip at all, e.g. see with clip_length and chunk_size parameters in our configs:
```
>>> chunk_size = 50
>>> clip_length = 150
>>> num_train_clips = 1
>>> list(np.arange(0, clip_length * (num_train_clips - 1), chunk_size, dtype=int))
[]
>>> list(np.arange(0, clip_length * num_train_clips, chunk_size, dtype=int))
[np.int64(0), np.int64(50), np.int64(100)]
>>> num_train_clips = 2
>>> list(np.arange(0, clip_length * (num_train_clips - 1), chunk_size, dtype=int))
[np.int64(0), np.int64(50), np.int64(100)]
>>> list(np.arange(0, clip_length * num_train_clips, chunk_size, dtype=int))
[np.int64(0), np.int64(50), np.int64(100), np.int64(150), np.int64(200), np.int64(250)]
```

Is this intended?

openretina/data_io/sridhar_2025/dataloaders.py
```diff
@@ -813,7 +815,7 @@ def __init__(
         num_of_frames: int = 15,
         num_of_layers: int = 1,
         device: str = "cpu",
-        time_chunk_size: Optional[int] = None,
+        time_chunk_size: int = 1,
```
thomasZen, line 814: setting it to None lead to bugs, it's set to this value in the config.

openretina/data_io/sridhar_2025/dataloaders.py
```diff
@@ -682,16 +682,18 @@ def frame_movie_loader(
...
+
             locations = get_locations_from_stas(
                 sta_dir=os.path.join(basepath, sta_dir),
                 retina_index=retina_index,
-                cells=[cell_index]
-                if cell_index is not None
-                else [
-                    x
-                    for x in range(0, train_responses.shape[0] + len(excluded_cells[retina_index]))
-                    if x not in excluded_cells[retina_index]
```
thomasZen, line 693: this would lead to an error if excluded_cells = None

openretina/data_io/sridhar_2025/dataloaders.py
```diff
@@ -470,7 +470,7 @@ def frame_movie_loader(
     crop: int | tuple[int, int, int, int] = 0,
     num_of_trials_to_use: int | None = None,
     start_using_trial: int = 0,
-    num_of_frames=None,
```
thomasZen, line 473: None would have resulted in an error below. I set the default to the value in the config.

openretina/data_io/base_dataloader.py
```diff
@@ -344,8 +344,10 @@ def get_movie_dataloader(
         ValueError:
             If `allow_over_boundaries` is False and `chunk_size` exceeds `scene_length` during training.
     """
-    if isinstance(responses, torch.Tensor) and bool(torch.isnan(responses).any()):
-        print("Nans in responses, skipping this dataloader")
+    if (isinstance(responses, torch.Tensor) and bool(torch.isnan(responses).any())) or (
+        isinstance(responses, np.ndarray) and bool(np.isnan(responses).any())
```
thomasZen, line 313: in case it's a numpy array, which we allow, it also makes sense to check against nans

openretina/eval/metrics.py
```diff
@@ -159,7 +159,7 @@ def MSE_stop3d(model: torch.nn.Module, loader, avg: bool = True, device: str = "
 
         ret = (output - target) ** 2
         if np.any(np.isnan(ret)):
-            warnings.warn(f"{np.isnan(ret).mean:.1%} NaNs")
+            warnings.warn(f"{np.isnan(ret).mean():.1%} NaNs")
```
thomasZen, line 162: we did not call the function.

openretina/models/linear_nonlinear.py
```diff
@@ -149,12 +149,12 @@ def __init__(
         self.location = rf_location
 
         regularizer_config_spat = (
-            dict(padding=laplace_padding, kernel=self.spat_kernel_size)
+            dict(padding=laplace_padding, kernel=spat_kernel_size)
```
thomasZen, line 154: `self.spat_kernel_size` was not defined.
I'm unsure why the type checker did not catch this

openretina/models/linear_nonlinear.py
```diff
@@ -149,12 +149,12 @@ def __init__(
...
             if smooth_regularizer_spat == "GaussianLaplaceL2"
             else dict(padding=laplace_padding)
         )
         regularizer_config_temp = (
-            dict(padding=laplace_padding, kernel=self.temp_kernel_size)
+            dict(padding=laplace_padding, kernel=in_shape[1])
```
thomasZen, line 159: self.temp_kernel_size was not defined.

openretina/modules/core/base_core.py
```diff
@@ -208,6 +208,9 @@ def group_sparsity_0(self) -> torch.Tensor:
         return torch.sum(torch.stack(result_array))
 
     def group_sparsity(self) -> torch.Tensor:
+        if len(self.features) <= 1:
```
thomasZen, line 211: if len(self.features) == 1, we get to `sparsities = []` (as `for feat in self.features[1:]`), and that leads to `torch.stack([])` which will fail.

openretina/modules/core/gru_core.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         batch_adaptation: bool = False,
         use_avg_reg: bool = False,
         nonlinearity: str = "ELU",
-        conv_type: str = "custom_separable",
+        convolution_type: str = "custom_separable",
```
thomasZen, line 44: config speaks of convolution_type, see:
https://github.com/open-retina/open-retina/blob/a6aae9240c587577dc1d8a8b60ad5b0eca441e8d/configs/model/base_core_readout.yaml#L15

this did not fail as we use **kwargs...

openretina/modules/core/gru_core.py
```diff
@@ -237,7 +237,7 @@ def group_sparsity(self):  # check if this is really what we want
         for layer in self.features:
             if hasattr(layer, "conv"):
                 spatial_weight_layer = layer.conv.weight_spatial
-                norm = spatial_weight_layer.pow(2).sum([2, 3, 4]).sqrt().sum(1)
+                norm = spatial_weight_layer.pow(2).sum([1, 2, 3, 4]).sqrt().sum(1)
                 sparsity_loss_layer = (spatial_weight_layer.pow(2).sum([2, 3, 4]).sqrt().sum(1) / norm).sum()
```
thomasZen, line 241: it seemed like `sparsity_loss_layer` was always constant as it divided `norm/norm` (check the definition of norm and the dividend, they were exactly the same I think.

openretina/modules/nonlinearities.py
```diff
@@ -24,4 +24,6 @@ def __init__(self, a=1.0, b=0.0, w=1.0, learn_a=False):
         self.w = nn.Parameter(torch.tensor(w), requires_grad=True)
 
     def forward(self, x):
-        return self.w * torch.log(1 + torch.exp(self.a * x + self.b))
+        scaled_x = self.a * x + self.b
+        # using torch's softplus is numerically more stable
```
thomasZen, line 28: see comment.

openretina/data_io/hoefling_2024/stimuli.py
thomasZen: Metrics before:
```
  Correlation                   : 0.5458
  MSE                           : 0.3823
  FEVe                          : 0.3947
  FEV (expl. var ratio)         : 0.4182
  Poisson loss                  : 0.8215
```

Metrics after:
```
  Correlation                   : 0.5591
  MSE                           : 0.3740
  FEVe                          : 0.4158
  FEV (expl. var ratio)         : 0.4182
  Poisson loss                  : 0.8184
```

thomasZen: It also seems like it's not possible to bleed into the val/test movie, as we split them apart before:
https://github.com/open-retina/open-retina/blob/27f7fd730c6f42b6918c3d4c8a12bf01f1c8e37f/openretina/data_io/base_dataloader.py#L116-L145

openretina/data_io/hoefling_2024/stimuli.py
```diff
@@ -104,7 +104,7 @@ def gen_start_indices(
                              108 training clips for the 20 different sequences
     :param val_clip_idx:     list of integers indicating the clips to be used
                              for validation
-    :param clip_length:      clip length in frames (5s*30frames/s = 150 frames)
+    :param clip_length:      clip length in  (5s*30frames/s = 150 frames)
```
fededagos: what happened here?

openretina/modules/core/gru_core.py
fededagos: Interesting... Maybe we should consider removing or adding a warning for the caught **kwargs

fededagos (approved):
These all make sense to me. Interesting that the type checker did not catch a few of them... Anyway, okay to merge, there is only one doc inconsistency but it is in a numpy docstring, I already fixed those in the docs PR so we can ignore it for now.

openretina/data_io/hoefling_2024/stimuli.py
thomasZen: fixed.

openretina/modules/core/gru_core.py
thomasZen: I'd vote to remove them, but we can do that in a separate PR I think

PR #180, opened 2026-02-27
Spatial Contrast (SC) model implementation

shashwatsridhar:
Hey everyone,

I apologize for taking my time with this, but now I have the SC model implemented in the openretina framework. These changes are largely independent of other files in the repo, save for renaming a couple of config files created by @Mvystrcilova for her models (primarily for filename consistency). The LNs and CNNs she implemented still run fine, and all existing tests pass without any errors. I haven't added any tests specific to this model, but if you all feel its necessary, it should be fairly easy to add some.

Keen to know what you all think, and happy I can contribute!


openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,292 @@
...
+    """Parametrized softplus: a * log(1 + exp(b*x + c)), with learnable a, b, c."""
+
+    def __init__(self, a: float = 1.0, b: float = 1.0, c: float = 0.0):
+        super().__init__()
+        self.a = nn.Parameter(torch.tensor(a, dtype=torch.float32), requires_grad=True)
+        self.b = nn.Parameter(torch.tensor(b, dtype=torch.float32), requires_grad=True)
+        self.c = nn.Parameter(torch.tensor(c, dtype=torch.float32), requires_grad=True)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.a * torch.log(1 + torch.exp(self.b * x + self.c))
```
thomasZen: I think this is the same as `ParametrizedSoftplus(w=a, a=b, b=c, learn_a=True)` defined here:
https://github.com/open-retina/open-retina/blob/15336fdcf7175ddfe33ea5fc0cfaf41da4f7f44f/openretina/modules/nonlinearities.py#L16-L29

The only difference is that ParametrizedSoftplus uses the pytorch implementation of Softplus instead of `torch.log(1 + torch.exp(x))` as it's numerically more stable for big input values (see: https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.softplus.html)

openretina/utils/sta_processing.py
```diff
@@ -0,0 +1,309 @@
...
+    temporal_variances = np.var(sta, axis=(1, 2))
+    peak_temporal_idx = np.argmax(temporal_variances)
+
+    spatial_frame = sta[peak_temporal_idx].copy()
+
+    gaussian_params = fit_2d_gaussian(spatial_frame)
+
+    # Flip polarity for OFF cells so spatial filter is always positive
+    if gaussian_params["amplitude"] < 0:
+        gaussian_params["amplitude"] = -gaussian_params["amplitude"]
```
thomasZen: nitpick, is this not the same as (which is a bit easier to read at least for me as it doesn't rely on the if statement):
```suggestion
    gaussian_params["amplitude"] = np.abs(gaussian_params["amplitude"])
```

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,292 @@
...
+        # target_spatial_shape ensures the STA is cropped to match the stimulus dimensions
+        target_spatial_shape = (in_shape[-2], in_shape[-1])
+        spatial_filter, temporal_filter, gaussian_params = load_sta_and_extract_filters(
+            sta_dir=sta_dir,
+            file_name=sta_file_name,
+            flip_sta=flip_sta,
+            target_spatial_shape=target_spatial_shape,
+            temporal_crop_frames=temporal_crop_frames,
+            sigma_contour=sigma_contour,
+        )
```
thomasZen, line 123: I think in `sta_dir` can be a remote file path pointing to huggingface, and `load_sta_and_extract_filters` uses np.load directly on it if I'm not mistaken. If so you can make sure that it's converted to a local path first with something like:

```suggestion
        local_sta_dir = get_local_file_path(sta_dir)
        spatial_filter, temporal_filter, gaussian_params = load_sta_and_extract_filters(
            sta_dir=local_sta_dir,
            file_name=sta_file_name,
            flip_sta=flip_sta,
            target_spatial_shape=target_spatial_shape,
            temporal_crop_frames=temporal_crop_frames,
            sigma_contour=sigma_contour,
        )
```

Be sure to `from openretina.utils.file_utils import get_local_file_path` at the top though.

openretina/models/spatial_contrast.py
thomasZen: Through AI feedback I noticed that it can also be a problem if we learn `a` in `self.a * torch.log(...)` you can potentially get a negative a during training, which will lead to negative predictive spiking rates and probably probably deteriorate training if using the Poisson loss.
(This is an potentially issue in the ParametrizedSoftplus implementation, too, though.)

Did you run into this issue before (i.e. getting a negative self.a and therefore negative predicted firing rates for a neuron)?

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,292 @@
...
+        filtered = F.conv1d(x_reshaped, kernel, padding=0)
+
+        return rearrange(filtered, "(b h w) 1 t -> b t h w", b=batch, h=h, w=w)
+
+    def compute_imean(
+        self, temp_filtered: Float[torch.Tensor, "batch time height width"]
+    ) -> Float[torch.Tensor, "batch time"]:
+        """Spatially-weighted mean: sum(spatial_filter * stimulus) over space."""
+        weighted = self.spatial_filter * temp_filtered
+        return weighted.sum(dim=(-2, -1))
```
thomasZen: The docstring says "mean", but the code says `weighted.sum`. Is this as the input is already normalized, or is this a mismatch between the docstring and the code?

So potentially it could make sense to do something like:
```suggestion
        """Spatially-weighted mean: sum(spatial_filter * stimulus) over space."""
        weighted = self.spatial_filter * temp_filtered
        normalizer = self.spatial_filter_sum + 1e-8
        return weighted.sum(dim=(-2, -1)) / normalizer
```

This is without knowing the model and implications though :sweat_smile:

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,292 @@
...
+
+        Output time dimension is reduced by (temporal_filter_length - 1).
+        """
+        batch, channels, time, h, w = x.shape
+
+        x_reshaped = rearrange(x, "b c t h w -> (b h w) c t")
+
+        # conv1d cross-correlates, which matches STA temporal order directly
+        # (index 0 = furthest past, index -1 = closest to spike)
+        kernel = self.temporal_filter.view(1, 1, -1)
```
thomasZen, line 211: this assumes there's 1 channel right?
If so and if the model does not support multiple channels, best state that in the docstring of the class and throw an exception if `in_shape: (channels, time, height, width)` passed to the constructor wants multiple channels.

thomasZen (commented):
Awesome, nice implementation!

I so far only commented and will also @fededagos a chance to look at the PR.

Regarding the suggestion to test different models on the same data set:
Do you think it's plausible to run the spatial contrast model for all cells of the dataset and adapting the implementation in the future roughly as follows (I can also help / work on that at some point if it makes sense):
- Have an option to return all cells of the dataset in the dataloader
- Adapt the model to deal with multiple cells an learn the 4 parameters for each cell individually (so you'd get `4 x numberOfCells`) params in the model
- If so, we could also integrate it into the core readout framework by having a pass-trough core, and treating the contrast model as a readout (I think the same applies to the linear_nonlinear model of Michaela)

openretina/models/spatial_contrast.py
thomasZen: We also integrated the data set that has UV/Green input channels from the eulerlab, if there's a clean path to making use of multiple channels it would likely be worth the effort, though (we support just averaging the color channels already in another code path, but maybe using both UV/Green individually makes sense, too)

openretina/models/spatial_contrast.py
shashwatsridhar: You're right on both fronts here, thanks for pointing those out. I didn't notice the existing softplus implementation, and it is indeed identical when instantiated the way you suggested, so I'll make this change. The point about the possibly negative a is also correct - it is something I accounted for in my published code, but overlooked when reimplementing here. I'll probably add a clamp on `a` (now a bit confusingly `w` in the existing softplus class) during the training step of the model. That should do it right?

openretina/utils/sta_processing.py
shashwatsridhar: Agreed! Will change.

openretina/models/spatial_contrast.py
shashwatsridhar: Ah wasn't aware that could be an issue! Will definitely fix.

openretina/models/spatial_contrast.py
shashwatsridhar: Ah whoops 😅 I did overlook this one, and didn't notice because the models still converged (likely because the filters are normalized). AI did the documentation for me, so it was no smarter than I haha... will fix!

openretina/models/spatial_contrast.py
shashwatsridhar: Yes, it supports just one channel, I'll update the docstring. I don't believe the extension to two or more channels is conceptually trivial. If we do compute the imean and local spatial contrast for different color channels, do we combine the contrast before combining with the lsc, or combine the weighted sum after? Should there be weights for each color channel, and if yes, what do they represent biologically? I think it is best if the model is restricted to a single channel (averaged across color channels if need be), but would certainly be an interesting avenue to explore in the future with this model.

openretina/models/spatial_contrast.py
shashwatsridhar: Looked into this a little, and turns out the better way is to override `optimizer_step` in the model class and add parameter clamping after `optimizer.step()` is called. [source](https://github.com/Lightning-AI/pytorch-lightning/issues/2382)

thomasZen: That sounds good.

openretina/models/spatial_contrast.py
thomasZen: Fair, makes sense. Best make sure that there's a early error if somebody tries to use the model with multiple channels.

configs/dataloader/sridhar_2025_nm_single_cell.yaml
```diff
@@ -12,42 +16,52 @@ fixation_files:
...
   "02": []
   "03": []
   "04": []
-retina_specific_crops: True
 hard_coded: null
 flip_imgs: False
 shuffle: null
 get_locations: True
-sta_dir: "stas"
+sta_dir: "stas"
```
thomasZen:
```suggestion
sta_dir: "stas"

```
It's slightly better to end a each file with a new line (e.g., see https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

configs/dataloader/sridhar_2025_wn_single_cell.yaml
```diff
@@ -21,33 +32,32 @@ test_image_path:
...
+  "01": []
+  "02": []
+  "03": []
+  "04": []
 extra_frame: 0
 hard_coded: null
 chunked_sampling: True
 get_locations: True
-sta_dir: "stas"
+sta_dir: "stas"
```
thomasZen:
```suggestion
sta_dir: "stas"

```

configs/dataloader/sridhar_nm_2025.yaml
```diff
@@ -45,7 +45,7 @@ device: 'cuda'
...
+    "01": [ 1, 5, 6, 7, 9, 10, 12, 14, 16, 19, 20, 22, 25, 26, 29, 30, 31, 32, 33, 34, 35, 43, 46, 53, 54, 57, 62, 63, 64, 70, 71, 72, 77, 78, 79, 81, 82, 83, 86, 89, 90, 91, 94, 96, 97, 101, 103, 106, 112, 114, 116, 121, 122, 126, 127, 128, 130, 133, 136, 140, 143, 144, 145, 146, 148, 149, 155, 159, 162, 168, 171, 176, 177, 178, 182, 185, 189, 191, 193, 197, 200, 209, 221, 222, 227, 230, 231, 234, 235, 236, 242, 243, 245, 247, 250, 254, 257, 258, 261, 262, 269, 271, 273, 274, 277, 281, 292, 296, 301, 303, 304, 307, 310, 311, 312, 313, 317, 318, 320, 325, 330, 331, 332, 333, 340, 341, 342, 344, 350, 352, 353, 354, 357, 358, 362, 363, 365 ]
```
thomasZen, line 50: how did you decide which cells are excluded? if it's easy to write down in a sentence, add a comment above.

configs/model/single_cell_sc.yaml
```diff
@@ -0,0 +1,30 @@
...
+# --- Temporal filter settings ---
+temporal_crop_frames: 30    # Keep this many of the most recent STA frames
+
+# --- Learnable parameter initial values ---
+# The model has only 4 learnable parameters: w, a, b, c
+learning_rate: 0.01
+w_init: 0.0     # Spatial contrast weight (0 = start as a pure LN model)
+a_init: 4.0     # Output scaling
+b_init: 1.0     # Input gain
+c_init: -5.0    # Offset (negative = low baseline firing rate)
```
thomasZen:
```suggestion
c_init: -5.0    # Offset (negative = low baseline firing rate)

```

configs/sridhar_2025_nm_sc.yaml
```diff
@@ -0,0 +1,53 @@
...
+  output_dir: ${hydra:runtime.output_dir}
+
+matmul_precision:
+  _target_: torch.set_float32_matmul_precision
+  precision: highest
+
+trainer:
+  max_epochs: 100
+  accumulate_grad_batches: 1
+  gradient_clip_val: 1
```
thomasZen:
```suggestion
  gradient_clip_val: 1

```

notebooks/training_sc_multicell_sridhar_2025.ipynb
```diff
@@ -0,0 +1,816 @@
...
...dO1dGjR/X6669r5syZjgsWAAAAAAAAAAAAqCSqxAjBHTt2aOPGjZKkcePGWSQDc02ePFnt2rWTJL355pvKyMiwa4wAAAAAAAAAAABAZVQlEoIrVqwwlh988MFC6zg5OemBBx6QJF25ckXh4eH2CA0AAAAAAAAAAACo1KpEQnDTpk2SJC8vL91www1W64WGhhrLmzdvtnlcAAAAAAAAAAAAQGVXJRKChw4dkiQFBgbKxcX6bQ+Dg4MLbAMAAAAAAAAAAADUZNaza5VEamqq4uLiJEn+/v5F1q1Tp468vLyUlJSkmJiYEu/jzJkzRa4/f/58idsCAAAAAAAAAAAAKpNKnxBMSEgwlr29vYutn5sQTExMLPE+AgICyhQbAAAAAAAAAAAAUNlV+ilDU1NTjWVXV9di67u5uUmSUlJSbBYTAAAAAAAAAAAAUFVU+hGC7u7uxnJ6enqx9dPS0iRJHh4eJd5HcdOLxsTE6KabbpJkOWIRAAA4TmZmZpH3FgaAqiIzM9NY5nYFAAAAAFBzNW7c2Ga/d1X6X9F8fHyM5ZJMA5qUlCSpZNOL5iru3oR5O+UffPBBidsFAAC283//93/F/g8HgKogNjbWWO7WrZsDIwEAAAAAOFJMTIzNfu+q9FOGuru7q169epKkM2fOFFn38uXLRkKQ+wICAAAAAAAAAAAAVWCEoCS1b99eGzduVFRUVJHTgx0+fNhYbteuXYXtv2PHjtqxY4eknOl8cqcP3bFjh5o0aVJh+0HFOn/+vHGFNceq8uN4VS0cr6qluh6vxo0bOzoEAKgQefsbDRo0qJTTIVfX/yWwHc4ZlBbnDEqLcwZlwXmD0uKcQWmV95yx5e9dla+nWYhevXpp48aNSkpK0u+//64bb7yx0HoRERHGcs+ePSts/+7u7goJCZFkOUqxSZMmTFVWRXCsqhaOV9XC8apaOF4AUPnk7W9UBfwvQWlxzqC0OGdQWpwzKAvOG5QW5wxKq7KdM5V+ylBJGj58uLH80UcfFVonOztbH3/8sSTJz89Pffr0sUdoAAAAAAAAAAAAQKVWJRKC3bp1U+/evSVJixcv1tatWwvUef3113Xo0CFJ0sSJE1WrVi27xggAAAAAAAAAAABURlViylBJevPNN9WzZ0+lpKRowIABeu6559SnTx+lpKToiy++0KJFiyRJQUFBmjx5soOjBQAAAA
```
thomasZen, line 439: We currently try to submit the non executed notebooks without data (I think @fededagos might have a way to show plots on the documentation website)

notebooks/training_sc_sridhar_2025.ipynb
```diff
@@ -0,0 +1,551 @@
...
...2VzBlc5iq4nlKHab0RL54nJiN5ctzlGw6tgW0eRfV8oQDXdzUkz1NhrjOs0fyUdeysUwcppL4x8kP6GYyISCE0O9/hbniT13Xl8YatYVsu4sj+TI/HOzSXpvUYJsmcqkSpRQ8UjdO2mV23BQJ6hSZcZhy9HFt2HS4KNRhateKfm82TQxK7+MVm8NUR8cMZTJLH8GnsRnMEsln1P9z+i6XEK4t983myD4uaMvxNwV2mOovtQimHvCAB+CBD3wgLrroojo+jmGYIlwFTBbf1MLWKMLh7TAtqgATy1AayWcZ5Bk7ShtYjGZibJFL2YH7mtzQB1xzCabmu/6Gw1TmbX2y2EQLAJGlkDaacbFByZAGe0yj+ISnbd8VUAtrTOihZm8pW9Gg6YVceyRf9bEUjcPlMAVAlSRocX60UpqzXyyYIvelEckHFkw1kchsR5HwITCNKE0cpqYvqCaSL14ALLeQmesw1dIFkWVCT6fpMLWzBccerC+UhtVhaiwLhc69eB5WTNGi+VK6kszisFJ6P20VTPkeZOb72lgvjeQjDlMhC6ZSDJF6ZpNFZgwq4enntiUOU419HkahMc9L3YyIkKLLkXzGBiQhLJt7yv/c5Bk9FcHofYtAPJdPj6HF43zbfTFve8pbTOriuCf9e5Wa1uWSpjJ5rgga+UpOROowNRFMuRymgLiWkY3kVFBz9bWu69vWuJ7snzNQ5Pm9eqR2/waIawUtvmV3jOGQjUwkH3GY0s5Lpubijqjq/oncHuvnbvdAfwY3zWWIzo9WAs9S463ziNrJbIIpPpFNgApbbIIpfwcbjV0/aotD08bQXi9vy/E3BdqulGpef89UQy2CqV27YkX/WWedVcfHMQxThGsXUherGA1k/3psr5/sbhKYCKZIJJ/NRpQ6TAkWTDUWOhgNPIGVYHpNd8sNvSCxemTxm3rlRPIZsXjZSL6CRXvb5HpWt5KI2FlLkdntNBEqJAhVRiSffly+5jBlDoGaXvCxnY9BD3eLuSL5ACAKqxdMZR2m4kg+ErVAFrLmLWYzFSPNgnkkAnhiGnsRpm4OyQsqiuSzOEwtei/nFUTGXCXdMbT/pdfLGsmXc57ZYWp2bM++YRgVnsM+PA+rpsg5YCkiATIGyM6NLEM7DHyhLTBrf1GyWcWI5BuCiTEj+abnMjsGVRB6P9jAcc9CseJ1Q9qgUtkxid7
```
thomasZen, line 237: potentially clear notebook.

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,469 @@
...
+            evaluation_loss=evaluation_loss,
+            data_info=data_info,
+        )
+
+    def optimizer_step(
+        self,
+        epoch: int,
+        batch_idx: int,
+        optimizer: Union[Optimizer, LightningOptimizer],
+        optimizer_closure: Optional[Callable[[], Any]] = None,
```
thomasZen:
```suggestion
        optimizer: Optimizer | LightningOptimizer,
        optimizer_closure: Callable[[], Any] | None = None,
```
Nitpick: You could also write above nowadays :)

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,469 @@
...
+    ) -> None:
+        """Override implementation to allow parameter clamping at each training step"""
+        optimizer.step(closure=optimizer_closure)
```
thomasZen, line 297: cool :)

openretina/utils/sta_processing.py
thomasZen: (in case you forgot to change and still want to, but feel free to keep it as is, too)

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,469 @@
...
+
+        # To avoid negative model predictions, clamp the gain parameter in the nonlinearity
+        self.nonlinearity.w.clamp(min=0.0)
```
thomasZen:
```suggestion
        self.nonlinearity.w.clamp_(min=0.0)
```
`clamp` is not inplace, so it will return a value but won't update w.

E.g.:
```
>>> import torch
>>> import numpy as np
>>> x = torch.Tensor(-np.ones(1))
>>> x.clamp(min=0.0)
tensor([0.])
>>> x
tensor([-1.])
>>> x.clamp_(min=0.0)
tensor([0.])
>>> x
tensor([0.])
```

thomasZen (approved):
Looks mostly good from my side, I approved it already.
I added smaller comments, and found a bug that is easy to fix (i.e. https://github.com/open-retina/open-retina/pull/180/changes#r2931401993).

Tagging @fededagos and @Mvystrcilova if they want to take a look, too, otherwise feel free to merge after checking the new comments.

openretina/models/spatial_contrast.py
shashwatsridhar: Ah thanks for this, I wasn't aware -- I didn't explicitly check if the clamping worked 🙈

shashwatsridhar:
Thanks a lot for reviewing @thomasZen :) I'm sorry I just pushed without writing anything here, I was a bit short on time with preparations for COSYNE! I made the major modifications you recommended (with a fair bit of Claude Code help), but I wasn't able to test the code on the naturalistic movie data (it works well on the white noise data). If you have a chance to test it (by even simply running the notebook) on the naturalistic movies, and if it works well for you, I'd say its mostly good from my end, except the small changes you suggested. Since I'm currently at COSYNE, it'll be difficult for me to get to those soon, but I'll give it a shot!

notebooks/training_sc_sridhar_2025.ipynb
shashwatsridhar: Will do!

shashwatsridhar:
All changes suggested above have been implemented. I don't have access to a GPU right away to test the NM stuff, might be able to do this in the next few days. But otherwise, its good from my end.

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,469 @@
...
+        epoch: int,
+        batch_idx: int,
+        optimizer: Union[Optimizer, LightningOptimizer],
+        optimizer_closure: Optional[Callable[[], Any]] = None,
+    ) -> None:
+        """Override implementation to allow parameter clamping at each training step"""
+        optimizer.step(closure=optimizer_closure)
+
+        # To avoid negative model predictions, clamp the gain parameter in the nonlinearity
+        self.nonlinearity.w.clamp_(min=0.0)
```
fededagos: This causes an autograd `RuntimeError` because it operates directly on a leaf tensor. Workaround is: 

```suggestion
        self.nonlinearity.w.data.clamp_(min=0.0)
```

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,469 @@
...
+        self.log("train_loss", loss, on_step=False, on_epoch=True)
+        self.log("train_correlation", correlation, on_step=False, on_epoch=True, prog_bar=True)
+
+        return loss
+
+    def validation_step(self, batch: tuple[str, DataPoint], batch_idx: int) -> torch.Tensor:
+        session_id, data_point = batch
+
+        model_output = self.forward(data_point.inputs, session_id)
+        loss = self.loss.forward(model_output, data_point.targets) / sum(model_output.shape)
```
fededagos: I think this loss computation might be wrong. I.e. `sum(model_output.shape)` will sum the shapes, but I guess your idea was to normalise by the total number of elements, i.e. `model_output.numel()`?
But I also notice that the training loss is not normalised, so I would suggest either dropping this here or adding it to the training so the loss magnitudes are comparable :)

openretina/models/spatial_contrast.py
```diff
@@ -0,0 +1,469 @@
...
+            self.rf_location = full_rf_location
+            self._crop_bounds = None
+
+        # Filters are registered as buffers: non-trainable but saved in state_dict
+        self.register_buffer("spatial_filter", torch.from_numpy(spatial_filter))
+        self.register_buffer("temporal_filter", torch.from_numpy(temporal_filter))
+        self.register_buffer("spatial_filter_sum", torch.tensor(spatial_filter.sum()))
+
+        # The 4 learnable parameters
+        self.nonlinearity = ParametrizedSoftplus(w=a_init, a=b_init, b=c_init, learn_a=True)
```
fededagos: It is really minor, and an internal thing, but it could be confusing how the param names rotate in here, because some refer to the SC model, and some to the softplus internals, i.e. b_init (SC) gets assigned to a of the softplus, and so on. Could honestly just leave as is, or you could explicitly use the parameters in the forward method like you are doing in `spatial_constrast_model_readout.py`

fededagos: This actually also causes an error in the notebook, because you are calling: 
```
print("Trained parameters:")
print(f"  w (spatial contrast weight): {model.w.data.item():.4f}")
print(f"  a (output scaling):          {model.nonlinearity.a.data.item():.4f}")
print(f"  b (input gain):              {model.nonlinearity.b.data.item():.4f}")
print(f"  c (offset):                  {model.nonlinearity.c.data.item():.4f}")
```
But the nonlinearity only has w, a and b as attributes

notebooks/training_sc_multicell_sridhar_2025.ipynb
fededagos, line 1: In the first cell there is a reference to one of your internal cluster paths that gets assigned to the cache directory. I would maybe just comment it out, as people might still want to change it so it is good to have how there

configs/sridhar_2025_wn_sc_multicell.yaml
```diff
@@ -0,0 +1,65 @@
...
+  retina_index: null
+
+# Keep data_io.sessions in sync with dataloader.retina_index so that only
+# the selected session's responses are loaded. If you want to train all
+# sessions, set dataloader.retina_index to null and comment out the following lines.
+# To train on a single session, set dataloader.retina_index to the session index (1, 2, 4)
+# and uncomment the following lines.
+#data_io:
+#  sessions:
+#    - ${dataloader.retina_index}
```
fededagos: This causes an error in the multicell notebook. We can either: have one session in dataloader.retina_index, and the same session in data_io (i.e. your uncommenting suggestion), or all sessions in dataloading, and all sessions in data_io. Right now, in data_io (because the default parent config is used), we would only load session ["04"], and none of the others, but then the dataloader is asking for them. 
Given also the content of the notebook, I think it might just make sense to allow training this on one session at a time only. The fix would be
```suggestion
  retina_index: "04"

# Keep data_io.sessions in sync with dataloader.retina_index so that only
# the selected session's responses are loaded.
data_io:
  sessions:
    - ${dataloader.retina_index}
```

fededagos (changes requested):
Hey! Finally also got to review this. Looks great, it is fantastic to have the model incorporated. 
The PR looks very nice, also very nicely documented throughout. 
I got to test the notebooks as well, and as a result I left a few minor comments, and a few ones that would need resolution so that the notebooks can run end-to-end. 
Thanks a lot for this!

configs/sridhar_2025_wn_sc_multicell.yaml
shashwatsridhar: I see. I figured having the option to train on all datasets with just a small change in the config is a good feature to have, but without a simple way of having the user select an arbitrary set of sessions (which is currently not possible due to the dataloading function, as far as I understand), its still an incomplete implementation. I've just taken your suggestion for now, and maybe we come back to expand this to an arbitrary set of sessions in the future. Thanks for pointing this out.

notebooks/training_sc_multicell_sridhar_2025.ipynb
shashwatsridhar: Whoops, didn't set it back to the generic path. Will fix.

openretina/models/spatial_contrast.py
shashwatsridhar: I understand this can be confusing, and the error definitely needs to go. I heeded your suggestion and the code will now just use `nn.functional.softplus` in the forward function directly. Thanks for catching the error!

openretina/models/spatial_contrast.py
shashwatsridhar: Good catch! I think normalizing would be good, so I'll just add the `.numel()` normalization to both training and validation steps.

openretina/models/spatial_contrast.py
shashwatsridhar: Glad you noticed this! I didn't catch this since I didn't have a chance to run the notebooks before pushing the last changes.

shashwatsridhar:
Thanks for the review @fededagos :) I have made the changes, but again, have no chance to check this on the GPU before next week, unfortunately! Lemme know if something doesn't run -- I can investigate and fix next week if its something involved. Glad this looks otherwise good to you, though!

fededagos (approved):

fededagos:
Thanks a lot @shashwatsridhar! I will merge it, very happy this is now integrated!!

PR #182, opened 2026-03-05
Fix lag computation for Sridhar 

fededagos:
Latest PR broke Sridhar evaluation.
In their case the test dataloader does not use full test chunk size, pre-computes lag, and has a few unused frames at the end. The lag computation used for the rest of the datasets was breaking this. 
I tried to fix it in a bit of a more general sense, in case any other dataloaders will use their pattern in the future.

### DIFF
--- a/openretina/cli/eval.py
+++ b/openretina/cli/eval.py
@@ -39,9 +39,9 @@ def evaluate_model(cfg: DictConfig) -> float:
     log.info("Logging full config:")
     log.info(OmegaConf.to_yaml(cfg))
 
-    if cfg.paths.cache_dir is None:
-        raise ValueError("Please provide a cache_dir for the data in the config file or as a command line argument.")
-    if cfg.evaluation.model_path is None:
+    if cfg.get("paths").get("cache_dir") is None:
+        raise ValueError("Please provide paths.cache_dir for the data in the config file or via the command line.")
+    if cfg.get("evaluation", {}).get("model_path") is None:
         raise ValueError("Please provide evaluation.model_path to define which model to test.")
 
     # Set cache folder
--- a/openretina/cli/eval.py
+++ b/openretina/cli/eval.py
@@ -89,12 +89,13 @@ def evaluate_model(cfg: DictConfig) -> float:
         dataset = dl.dataset
 
         with torch.no_grad():
-            model_responses_torch_array: list[torch.Tensor] = []
+            model_responses_torch_array = []
             targets_array = []
             for data_point in dl:
                 model_resp_batch = model.forward(data_point[0].to(device), data_key=session)
-                model_responses_torch_array.extend(model_resp_batch.detach())
-                targets_array.extend(data_point[1])
+                model_resp = model_resp_batch.flatten(0, 1)
+                model_responses_torch_array.append(model_resp)
+                targets_array.append(data_point[1].flatten(0, 1))
             model_responses_torch = torch.concat(model_responses_torch_array)
             targets = torch.concat(targets_array).to(device)
             poisson_loss_session = poisson_loss(model_responses_torch.unsqueeze(0), targets.unsqueeze(0))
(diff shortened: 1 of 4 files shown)


openretina/utils/eval_utils.py
```diff
@@ -254,3 +258,66 @@ def print_report(self) -> None:
             print(f"  {'MSE':30s}: {self.mse_by_trial_mean:.4f} (+/-{self.mse_by_trial_std:.4f})")
 
         print("=" * 80)
+
+
+def align_responses_to_model_output(
+    targets: torch.Tensor,
```
thomasZen: a bit nitpicky, but I think it'd be more consistent if everything is a np.ndarray here.

openretina/utils/eval_utils.py
```diff
@@ -254,3 +258,66 @@ def print_report(self) -> None:
...
+        responses_by_trial = responses_by_trial[new_lag : new_lag + model_len]
+
+    if lag < 0:
+        lag = new_lag
+    elif new_lag != lag:
+        log.error(
+            f"Inconsistent lag between sessions: {new_lag=} {lag=}"
+            "\nThis might indicate a problem with the model or the data."
+        )
+        lag = new_lag
```
thomasZen: I'd prefer to throw an exception here. I could otherwise see it happening that people just ignore this log message and report the metric the get out eventually.

openretina/utils/eval_utils.py
```diff
@@ -254,3 +258,66 @@ def print_report(self) -> None:
...
+    Returns:
+        avg_responses: Aligned average responses.
+        responses_by_trial: Aligned responses by trial.
+        lag: Re-computed temporal lag of the model output with respect to the targets for this session's responses.
+    """
+    dataloader_target_len = targets.shape[0]
+    model_len = model_responses.shape[0]
+    full_len = avg_responses.shape[0]
+
+    if dataloader_target_len == model_len and full_len >= model_len:
```
thomasZen: if ```full_len == model_len``` and we do not have a frame_overhead, it would raise.
Shouldn't this be valid?

openretina/cli/eval.py
```diff
@@ -127,20 +127,16 @@ def evaluate_model(cfg: DictConfig) -> float:
...
+            model_responses=model_responses,
+            avg_responses=avg_responses,
+            responses_by_trial=responses_by_trial,
+            dataset=dataset,
+            lag=lag,
+        )
 
-        avg_responses = avg_responses[lag:]
         n_neurons_session = avg_responses.shape[1]
-        responses_by_trial = responses_by_trial[lag:]
```
thomasZen, line 143: I noticed later on we call jacknife as follows:
```
jackknife, _ = oracle_corr_jackknife(responses_by_trial, cut_first_n_frames=lag)
```

This leads to cutting the responses_by_trial twice, I think. We should likely pass 0 as the lag there?

openretina/utils/eval_utils.py
```diff
@@ -254,3 +258,66 @@ def print_report(self) -> None:
...
+                "dataloader pre-trims responses."
+            )
+        start_trim = getattr(dataset, "frame_overhead", 0) + getattr(dataset, "lag", 0)
+        new_lag = start_trim
+        avg_responses = avg_responses[start_trim : start_trim + model_len]
+        responses_by_trial = responses_by_trial[start_trim : start_trim + model_len]
+
+    else:
+        # Standard case: model output is shorter than full responses by lag in dataloader targets
+        new_lag = dataloader_target_len - model_len
```
thomasZen, line 317: throw an error if `new_lag < 0` or add an assert. This shouldn't happen, but if it happens it would throw off the numpy indexing.

openretina/cli/eval.py
thomasZen: (this problem was there before already, but I guess fixing it is worth it).

thomasZen (commented):
Was the PR that broke evaluation for the sridhar dataset https://github.com/open-retina/open-retina/pull/179?

I commented so far and I think it's worth and easy to improve evaluation a bit more (especially the jacknife oracle), but apart from that looks good. I would want to look at it once more afterwards, though.

openretina/utils/eval_utils.py
thomasZen: added value error.

thomasZen (approved):
Looks good.

PR #184, opened 2026-03-20
Add readout visualization for all readout layers

thomasZen:
Previous, the example to plot the readout weights mentioned in the [documentation](https://open-retina.org/package_docs/insilico/weight_visualisations/) did not work for "hoefling_2024_low_res" as the PointGaussianReadout did not implement the plotting function.

Now it plots the following for the "hoefling_2024_low_res" model:
readout_weight_hoefling_2024_low_res

I also made the Readout class an abstract class which enforces that certain methods have to be implemented in all subclasses. I implemented the feature visualization for all readout subclasses, too.

Here's how it looks for other models:

## GaussianMaskReadout

base_readout

## FactorizedReadout (randomly initialized with unusual params)

Figure_2

## LNPReadout (randomly initialized)

Figure_3

### DIFF
--- a/openretina/modules/readout/base.py
+++ b/openretina/modules/readout/base.py
@@ -3,15 +3,18 @@ Adapted from neuralpredictors:
 https://github.com/sinzlab/neuralpredictors/blob/v0.3.0.pre/neuralpredictors/layers/readouts/base.py
 """
 
+import os
 import warnings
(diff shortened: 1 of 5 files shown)


fededagos (approved):
Looks good! The refactor to `ABC` and enforcing the methods is also good improvement, really completes the readout refactor and "unification" now I think

PR #187, opened 2026-07-14
Debug/laplace1d

SamuelSuhai:
I think the following is a bug: The `Laplace1d` method convolved the input with a 1d laplace kernel and applied an aggregation function (mean or sum), so its outputs could be negative or positive. In `SimpleCoreWrapper` and `SingleCellSeparatedLNP` however it is used as a regularizer so I think it may not make sense to allow negative values. 

I applied the following changes: 
1. `Laplace1d` is now a filtering class (I removed the aggregation function) just like `Laplace` for 2d inputs. 
2. The regularizer class `TimeLaplaceL23dnorm`. then uses the filtering class  `Laplace1d`  just  like `FlatLaplaceL23dnorm` then use the `Laplace` class, by squaring components and applying some aggregation function. 
3. The places where `Laplace1d` was instantiated in models now instantiate `TimeLaplaceL23dnorm`.
4. When `TimeLaplaceL23dnorm` is called, then the weights of the temporal kernels are no longer reshaped, as this is done in the forward method of `TimeLaplaceL23dnorm` itself. 
5. Added a test of the  `SingleCellSeparatedLNP` and removed a default argument value that raised an error. 

Note that one alternative would be to just take the absolute of the convolution result in the forward method (as done [here in neuralpredictors](https://github.com/open-retina/open-retina/blob/4a54141e234ec5bc6fedee43ff2b57350ae17554/openretina/modules/layers/regularizers.py#L126) but I thought this is cleaner.

### DIFF
--- a/configs/model/single_cell_lnp.yaml
+++ b/configs/model/single_cell_lnp.yaml
@@ -10,7 +10,7 @@ smooth_weight_spat: 1.0
 smooth_weight_temp: 1.0
 sparse_weight: 1.0
 smooth_regularizer_spat: "LaplaceL2norm"
-smooth_regularizer_temp: "Laplace1d"
+smooth_regularizer_temp: "TimeLaplaceL23dnorm"
 laplace_padding: null
 fit_gaussian: false
 normalize_weights: false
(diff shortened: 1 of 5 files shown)


thomasZen (approved):
Looks good to me.

I added a small commit (https://github.com/open-retina/open-retina/pull/187/changes/bcce5c9db423b4864d167c1e6e5b10dbe082ea51) to make the test a bit more stringent and to update the config for the single cell LNP appropriately.  Best check this small commit by me and then feel free to merge.

PR #188, opened 2026-07-23
Small Bugfixes: FactorizedReadout mask regularization loss, response & stimulus rate info in maheswaranathan config

thomasZen:
## Sum(model_output.shape) vs model_output.numel()

- model_output.numel() = the product of the dimensions = the total number of elements = 32 × 50 × 100 = 160,000.
- sum(model_output.shape) = the sum of the dimension sizes = 32 + 50 + 100 = 182.

This only affects how we log the learning rate during validation.

## Other changes
- Normalization changes for the sparse autoencoder (code we didn't use in the paper)
- Update openretina version to the most recent one (1.1.1) in uv.lock
- Consistency change for some regularizers
- Fixed rate for stimulus and response in maheswaranathan configs (30Hz -> 100Hz according to the loaded data which states dt of 10ms)

### DIFF
--- a/openretina/modules/readout/factorized.py
+++ b/openretina/modules/readout/factorized.py
@@ -159,7 +159,7 @@ class FactorizedReadout(Readout):
     def regularizer(self, reduction: Optional[Literal["sum", "mean"]] = None) -> torch.Tensor:
         mask_r = self.reg[0] * torch.mean(torch.sum(torch.abs(self.mask_weights), dim=0))
         wt_r = self.reg[1] * torch.mean(torch.sum(torch.abs(self.feature_weights), dim=0))
-        reshaped_masked_weights = self.mask_weights.T.reshape(-1, 1, 1, self.mask_size[0], self.mask_size[1])
+        reshaped_masked_weights = self.mask_weights.reshape(-1, 1, 1, self.mask_size[0], self.mask_size[1])
         laplace_mask_r = self.reg[2] * self._input_weights_regularizer_spatial(reshaped_masked_weights, avg=False)
         if reduction == "mean":
             return mask_r + wt_r + laplace_mask_r / 3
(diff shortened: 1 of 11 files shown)


openretina/modules/readout/factorized.py
```diff
@@ -159,7 +159,7 @@ def forward(self, x: torch.Tensor, **kwargs: Any) -> torch.Tensor:
     def regularizer(self, reduction: Optional[Literal["sum", "mean"]] = None) -> torch.Tensor:
         mask_r = self.reg[0] * torch.mean(torch.sum(torch.abs(self.mask_weights), dim=0))
         wt_r = self.reg[1] * torch.mean(torch.sum(torch.abs(self.feature_weights), dim=0))
-        reshaped_masked_weights = self.mask_weights.reshape(-1, 1, 1, self.mask_size[0], self.mask_size[1])
+        reshaped_masked_weights = self.mask_weights.T.reshape(-1, 1, 1, self.mask_size[0], self.mask_size[1])
```
thomasZen, line 162: Factorized readout was only used in configs/model/core_klindt_readout, not in the default model config.

uv.lock
```diff
@@ -3374,7 +3374,7 @@ wheels = [
 
 [[package]]
 name = "openretina"
-version = "1.0.0"
+version = "1.1.1"
```
thomasZen, line 3377: we are at version 1.1.1 now already.

SamuelSuhai (approved):
Looks good.

openretina/modules/readout/factorized.py
thomasZen: self.masks has the shape (num_pixels, num_neurons). Above command garbled the neurons and the pixels in a bad way, this is fixed by transposing the 2D matrix.

PR #190, opened 2026-07-30
test: all losses should be invariant to batch size, time, neurons

SamuelSuhai:
Added a test to see if the loss and regularization metrics are invariant to batch size, time and neuron number. Currently they are not. This can be addressed in the future.

We currently run one session per forward path, sum the poisson/likelihood loss (using the default avg=False), and use potentially different aggregation functions for the regularization.

This has the following effects:
- We have to retune the hyperparameters for interpolations the regularization losses whenever we change the batch size and the number of neurons
- The likelihood loss is most interpretable if it shows the average likelihood per neuron, which it currently does not (as it is currently showing the sum)
- For example: if we increase the batch size this would increase the poisson loss, but not the readout regularization loss, so we end up regularizing less. 
- If we used avg = True in `PoissonLoss3d` then one session with many neurons would have the same weight as another session with only few neurons.

### DIFF
--- a/tests/models/test_core_readout_loss_scale_invariance.py
+++ b/tests/models/test_core_readout_loss_scale_invariance.py
@@ -1,0 +1,149 @@
+import pytest
+import torch
+
+from openretina.models.core_readout import BaseCoreReadout
+from openretina.modules.core.base_core import SimpleCoreWrapper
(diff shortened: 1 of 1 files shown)


thomasZen (approved):
Looks good to me.

PR #194, opened 2026-08-20
Support newest torch and mlflow versions and upgrade dependencies

thomasZen:
Newer torch versions require us to specify `weights_only=False`, also when selecting the best checkpoint with lightning. As lightning only supports weights_only in version 2.6, we have to require that version in our dependencies.

The newest version of mlflow wants a sqlite database as its storage, which we define now in the mflow config.

I also upgraded uv.lock for all dependencies, I think it's a good opportunity to do this now.


thomasZen:
https://gin.g-node.org/ is slow again, which leads to some notebooks not working.

PR #195, opened 2026-08-21
Improvements for sridhar dataloader

thomasZen:
- Fix trial selection, indexing, and extra_frame boundaries (these argument values were not used in the configs, but would have let to failures)
- Use correct per-retina fixations and load only selected retinas (there was a TODO comment for this)
- Prevent frame-cache reuse across trials (potentially affect 1% of training samples)
- Improve deterministic sampling, caching, rendering speed, and memory use (now dataloading is at ~2 min instead of ~10 min on the CIN cluster)
- Tighten defaults, typing, and validation.
- Used config with hyperparameter tuned config that matches the uploaded checkpoint.

## Evaluation on main did not work with old config
Also on main the command:
```
openretina eval --config-name=vystrcilova_2024_nm_cnn evaluation.model_path=sridhar_2025 evaluation.comp
ute_dataset_statistics=false +evaluation=default
```
did not work as the config was changed in https://github.com/open-retina/open-retina/pull/180 and did not match the uploaded model anymore (and the lag computation was off):
```
  File "/gpfs01/euler/User/tzenkel/laptop_mount/open-retina/.venv/lib64/python3.12/site-packages/torch/nn/modules/module.py", line 1789, in _call_
impl                                                                                                                                              
    return forward_call(*args, **kwargs)                                                                                                          
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                          
  File "/gpfs01/euler/User/tzenkel/laptop_mount/open-retina/openretina/modules/losses/poisson.py", line 13, in forward                            
    loss = output - target[:, lag:, :] * torch.log(output + self.bias)                                                                            
                    ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                            
RuntimeError: The size of tensor a (233) must match the size of tensor b (235) at non-singleton dimension 2
```

With the new config I confirmed that we get the same results on the uploaded checkpoint that we reported in the Elife paper.


thomasZen:
Jupyter tests failed due to gin outages during loading the ensemble model.

PR #197, opened 2026-09-17
General fixes

lhoefling:
Fixes a couple of existing small bugs and inefficiencies that surfaced during Qiu dataset integration. Changes include elaborate explanations of why they came about by Claude - I left them in for you to read, but we may want to delete them once read to keep code files clean and concise. I have not checked the tests so far.


openretina/callbacks/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Lightning callbacks used by the training CLI.
```
thomasZen, line 1: Good catch. I think we could delete the comment completely or at least everything except the first line, but I do not have a strong opinion on it.

thomasZen (approved):
Looks good, sometimes it might make sense to strip down the comments a bit.

